PR #5, opened 2023-06-16
implement loggingReconciler to factorize code

hervenicol:
This PR implements a new `loggingReconciler` which act as an abstraction layer between actual reconciled object (Cluster api CR, Service CR) and the core logic of the operator which only uses the object metadata.

This provides some benefits:
- Avoid code duplication
- Unified logic for all controllers


TheoBrigitte:
Handling the not found error in promtail-wiring.ReconcileDelete methods fixes the above error.

Yet with this approach we need to find a workaround to enable logging on vintage MC.

hervenicol:
Discussed with @TheoBrigitte , no crazy logic here, just some rewritting of some stuff nobody read before, so safe to merge.

hervenicol:
except it seems I can't bypass reviewer approval :sweat_smile:

QuentinBisson:
So would it not have been easier to create a fake ClusterCR for the vintage installations instead of having this abstraction?

hervenicol:
> So would it not have been easier to create a fake ClusterCR for the vintage installations instead of having this abstraction?

You mean, have a reconciler that creates / deletes a fake ClusterCR, then rely on it with the standard clusterReconciler? :thinking: 
That's an idea!

QuentinBisson:
Yes that way we would only interact with Cluster CR and not lose that information to metav1.Object. That would reduce casting of go types

hervenicol:
> Yes that way we would only interact with Cluster CR and not lose that information to metav1.Object. That would reduce casting of go types

We must ensure this extra cluster CR is not read by cluster-operator or any other unexpected tool that would consider it a WC though.

QuentinBisson:
I meant to fake on in the code but maybe we can discuss tomorrow :)

TheoBrigitte:
> I meant to fake on in the code but maybe we can discuss tomorrow :)

We had some thoughts around this.

We need 2 things:
- a common cluster interface which defines how to get low level information like: get the logging label value, get the app namespace, etc. We can later extend this interface with any new needed method. We have different implementation for MC vintage, WC vintage, and CAPI
- a fake vintage MC object with artificial information injected into it


Let's merge this PR as is, since it was mainly focused on `LoggingReconciler` abstraction and move those ideas into a separate PR.

QuentinBisson (approved):

PR #7, opened 2023-06-19
Implement logged-cluster interface

hervenicol:
Implement a `logged-cluster` interface with 3 implementations:
- capicluster
- vintagewc
- vintagemc

And let's try to keep the minimum amount of code in these, to limit specific code.

### DIFF
--- a/pkg/capicluster/cluster.go
+++ b/pkg/capicluster/cluster.go
@@ -1,6 +1,8 @@
 package capicluster
 
 import (
+	"fmt"
+
 	"github.com/giantswarm/logging-operator/pkg/key"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -19,4 +21,8 @@ func (o Object) GetLoggingLabel() string {
 
 func (o Object) GetAppsNamespace() string {
 	return o.Object.GetNamespace()
+}
+
+func (o Object) GetAppName(app string) string {
+	return fmt.Sprintf("%s-%s", o.GetName(), app)
 }
--- a/pkg/capicluster/cluster.go
+++ b/pkg/capicluster/cluster.go
@@ -23,6 +23,6 @@ func (o Object) GetAppsNamespace() string {
 	return o.Object.GetNamespace()
 }
 
-func (o Object) GetAppName(app string) string {
+func (o Object) AppConfigName(app string) string {
 	return fmt.Sprintf("%s-%s", o.GetName(), app)
 }
--- a/pkg/capicluster/cluster.go
+++ b/pkg/capicluster/cluster.go
@@ -1,0 +1,22 @@
+package capicluster
+
+import (
+	"github.com/giantswarm/logging-operator/pkg/key"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Object struct {
+	client.Object
+}
+
+func (o Object) GetLoggingLabel() string {
+	labels := o.Object.GetLabels()
+
+	value := labels[key.LoggingLabel]
+
+	return value
+}
+
+func (o Object) GetAppsNamespace() string {
+	return o.Object.GetNamespace()
+}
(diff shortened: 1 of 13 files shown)


pkg/capicluster/cluster.go
```diff
@@ -0,0 +1,28 @@
...
+type Object struct {
+	client.Object
+}
+
+func (o Object) GetLoggingLabel() string {
+	labels := o.Object.GetLabels()
+
+	value := labels[key.LoggingLabel]
+
+	return value
```
QuentinBisson, line 19: Maybe this function should return a boolean instead and not the whole label?

hervenicol: You know what? It was actually hard to write it returning a string, because it would make so much more sense for it to return a boolean :rofl: 
But we were so focused on "return the contents of the label" that we persisted in returning a string.

WDYT @TheoBrigitte , do you see some reasons not to return a boolean?

hervenicol: We discussed we keep it as is for the moment.

TheoBrigitte: We talked, let's keep it like this for now.

pkg/logging-reconciler/logging_reconciler.go
```diff
@@ -22,22 +23,21 @@ type LoggingReconciler struct {
...
 }
 
-func (l *LoggingReconciler) Reconcile(ctx context.Context, object client.Object) (result ctrl.Result, err error) {
+func (l *LoggingReconciler) Reconcile(ctx context.Context, lc loggedcluster.Interface) (result ctrl.Result, err error) {
 	// Logging should be enabled when all conditions are met:
-	//   - logging label is set and true on the object
-	//   - object is not being deleted
+	//   - logging label is set and true on the lc
+	//   - lc is not being deleted
 	//   - TODO(theo) global logging flag is enabled
```
TheoBrigitte, line 30: Maybe move this comment to where the declaration of common.IsLoggingEnabled is.

pkg/resource/promtail-toggle/observability_bundle_configmap.go
```diff
@@ -19,16 +17,16 @@ type app struct {
 }
 
 // ObservabilityBundleConfigMapMeta returns metadata for the observability bundle user value configmap.
-func ObservabilityBundleConfigMapMeta(object client.Object) metav1.ObjectMeta {
+func ObservabilityBundleConfigMapMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
 	return metav1.ObjectMeta{
-		Name:      fmt.Sprintf("%s-observability-bundle-user-values", object.GetName()),
-		Namespace: object.GetName(),
+		Name:      lc.GetAppName("observability-bundle-user-values"),
```
TheoBrigitte, line 22: This is not an app name. It is just a fixed name we use to create a configmap used a user value in the observability bundle app. Maybe we should have a fix name like `observability-bundle-logging-operator-user-values` ?

TheoBrigitte: Alright make sense to keep better names for uniqueness.

Or maybe `AppUserValueName` ? I have no strong opinion

TheoBrigitte:
> Remember that we need to use extra config and not user values hère so we could have observability-bundle-logging-config ?

How about we merge this and do the extra config change in another PR. :policeman:  Keep-PR-small police :policeman:

QuentinBisson:
You do you, but then maybe we need to have smaller tasks in the board

pkg/logging-reconciler/logging_reconciler.go
hervenicol: oops, done!

pkg/resource/promtail-toggle/observability_bundle_configmap.go
TheoBrigitte: Fine with me then :)

TheoBrigitte (approved):

PR #8, opened 2023-06-21
add --vintage toggle

hervenicol:
We need a different set of reconcilers running on vintage and CAPI clusters.
So we'll have a CLI flag to select which mode we want when running the logging-operator.


TheoBrigitte (commented):
This implementation looks alright, but I do have some concerns around integration with our eco system.

How do we choose to enable this flag or not ? From my understanding this flag should be controlled by values within the config repository, which then comes to:
- This value should also be read from a configuration file.
- Should we have a boolean here or rely on the provider value and make a choice internally ? A boolean enforces us to define this value for every installation, while provider value already exists.

QuentinBisson:
@TheoBrigitte what prevents us from adding this argument in the deployment if this is a vintage installation?

QuentinBisson (approved):

PR #10, opened 2023-06-23
add support for extravalues where supported

hervenicol:

We should not overwrite user config for WCs, as some customers set some custom ones.
So we decided to go for extra-configs.

But when discussing with @QuentinBisson yesterday, we understood that:
- vintage MCs don't support extraconfig because they're not reconciled by cluster-operator
- for capi clusters we're not sure

So I did this dirty thing that will work at least on vintage.
To be discussed.


QuentinBisson (approved):
Not blocking for now

TheoBrigitte (approved):

PR #11, opened 2023-06-23
Create logging-credentials secret

hervenicol:

Create a "source of trust" secret containing logging credentials:
- read user
- write user
- read password
- write password

Generates password when they are not set.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -34,7 +34,6 @@ import (
 
 	appv1 "github.com/giantswarm/apiextensions-application/api/v1alpha1"
 	"github.com/giantswarm/logging-operator/internal/controller"
-	"github.com/giantswarm/logging-operator/pkg/common"
 	loggingreconciler "github.com/giantswarm/logging-operator/pkg/logging-reconciler"
 	"github.com/giantswarm/logging-operator/pkg/reconciler"
 	loggingcredentials "github.com/giantswarm/logging-operator/pkg/resource/logging-credentials"
@@ -59,16 +58,12 @@ func init() {
 func main() {
 	var enableLeaderElection bool
 	var metricsAddr string
-	var credentialsSecretName string
-	var credentialsSecretNamespace string
 	var probeAddr string
 	var vintageMode bool
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
-	flag.StringVar(&credentialsSecretName, "credentials-secret-name", "logging-credentials", "The name for the secret the operator will use to store credentials.")
-	flag.StringVar(&credentialsSecretNamespace, "credentials-secret-namespace", "monitoring", "The namespace for the secret the operator will use to store credentials.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&vintageMode, "vintage", false, "Reconcile resources on a Vintage installation")
 	opts := zap.Options{
@@ -103,11 +98,6 @@ func main() {
 		os.Exit(1)
 	}
 
-	loggingConfig := common.LoggingConfig{
-		CredentialsSecretName:      credentialsSecretName,
-		CredentialsSecretNamespace: credentialsSecretNamespace,
-	}
-
 	promtailReconciler := promtailtoggle.Reconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
@@ -119,9 +109,8 @@ func main() {
 	}
 
 	loggingSecrets := loggingcredentials.Reconciler{
-		Client:        mgr.GetClient(),
-		Scheme:        mgr.GetScheme(),
-		LoggingConfig: loggingConfig,
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
 	}
 
 	loggingReconciler := loggingreconciler.LoggingReconciler{
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -110,7 +110,6 @@ func main() {
 
 	loggingSecrets := loggingcredentials.Reconciler{
 		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
 	}
 
 	loggingReconciler := loggingreconciler.LoggingReconciler{
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -36,6 +36,7 @@ import (
 	"github.com/giantswarm/logging-operator/internal/controller"
 	loggingreconciler "github.com/giantswarm/logging-operator/pkg/logging-reconciler"
 	"github.com/giantswarm/logging-operator/pkg/reconciler"
+	loggingcredentials "github.com/giantswarm/logging-operator/pkg/resource/logging-credentials"
 	promtailtoggle "github.com/giantswarm/logging-operator/pkg/resource/promtail-toggle"
 	promtailwiring "github.com/giantswarm/logging-operator/pkg/resource/promtail-wiring"
 	//+kubebuilder:scaffold:imports
@@ -107,12 +108,18 @@ func main() {
 		Scheme: mgr.GetScheme(),
 	}
 
+	loggingSecrets := loggingcredentials.Reconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
+	}
+
 	loggingReconciler := loggingreconciler.LoggingReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
 		Reconcilers: []reconciler.Interface{
 			&promtailReconciler,
 			&promtailWiring,
+			&loggingSecrets,
 		},
 	}
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -34,6 +34,7 @@ import (
 
 	appv1 "github.com/giantswarm/apiextensions-application/api/v1alpha1"
 	"github.com/giantswarm/logging-operator/internal/controller"
+	"github.com/giantswarm/logging-operator/pkg/common"
 	loggingreconciler "github.com/giantswarm/logging-operator/pkg/logging-reconciler"
 	"github.com/giantswarm/logging-operator/pkg/reconciler"
 	loggingcredentials "github.com/giantswarm/logging-operator/pkg/resource/logging-credentials"
@@ -56,15 +57,19 @@ func init() {
 }
 
 func main() {
-	var metricsAddr string
 	var enableLeaderElection bool
+	var metricsAddr string
+	var credentialsSecretName string
+	var credentialsSecretNamespace string
 	var probeAddr string
 	var vintageMode bool
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
-	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&credentialsSecretName, "credentials-secret-name", "logging-credentials", "The name for the secret the operator will use to store credentials.")
+	flag.StringVar(&credentialsSecretNamespace, "credentials-secret-namespace", "monitoring", "The namespace for the secret the operator will use to store credentials.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&vintageMode, "vintage", false, "Reconcile resources on a Vintage installation")
 	opts := zap.Options{
 		Development: true,
@@ -98,6 +103,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	loggingConfig := common.LoggingConfig{
+		CredentialsSecretName:      credentialsSecretName,
+		CredentialsSecretNamespace: credentialsSecretNamespace,
+	}
+
 	promtailReconciler := promtailtoggle.Reconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
@@ -109,8 +119,9 @@ func main() {
 	}
 
 	loggingSecrets := loggingcredentials.Reconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
+		Client:        mgr.GetClient(),
+		Scheme:        mgr.GetScheme(),
+		LoggingConfig: loggingConfig,
 	}
 
 	loggingReconciler := loggingreconciler.LoggingReconciler{
--- a/pkg/resource/logging-credentials/logging_operator_secrets.go
+++ b/pkg/resource/logging-credentials/logging_operator_secrets.go
@@ -39,14 +39,14 @@ func genPassword() string {
 
 // GenerateObservabilityBundleConfigMap returns a configmap for
 // the observabilitybundle application to enable promtail.
-func GenerateLoggingCredentialsBasicSecret(lc loggedcluster.Interface) (v1.Secret, error) {
+func GenerateLoggingCredentialsBasicSecret(lc loggedcluster.Interface) *v1.Secret {
 
 	secret := v1.Secret{
 		ObjectMeta: LoggingCredentialsSecretMeta(lc),
 		Data:       map[string][]byte{},
 	}
 
-	return secret, nil
+	return &secret
 }
 
 // Update a LoggingCredentials secret if needed
--- a/pkg/resource/logging-credentials/logging_operator_secrets.go
+++ b/pkg/resource/logging-credentials/logging_operator_secrets.go
@@ -1,0 +1,50 @@
+package loggingcredentials
+
+import (
+	"math/rand"
+	"strings"
+	"time"
+
+	loggedcluster "github.com/giantswarm/logging-operator/pkg/logged-cluster"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// LoggingCredentialsSecretMeta returns metadata for the logging-operator credentials secret.
+func LoggingCredentialsSecretMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
+	return metav1.ObjectMeta{
+		Name:      lc.AppConfigName("logging-credentials"),
+		Namespace: "monitoring",
+	}
+}
+
+// Generate a random 20-characters password
+func genPassword() string {
+	const length = 20
+
+	rand.Seed(time.Now().UnixNano())
+	chars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
+		"abcdefghijklmnopqrstuvwxyz" +
+		"0123456789" +
+		"~=+%^*/()[]{}/!@#$?|")
+
+	var b strings.Builder
+	for i := 0; i < length; i++ {
+		b.WriteRune(chars[rand.Intn(len(chars))])
+
+	}
+	str := b.String() // E.g. "ExcbsVQs"
+	return str
+}
+
+// GenerateObservabilityBundleConfigMap returns a configmap for
+// the observabilitybundle application to enable promtail.
+func GenerateLoggingCredentialsBasicSecret(lc loggedcluster.Interface) (v1.Secret, error) {
+
+	secret := v1.Secret{
+		ObjectMeta: LoggingCredentialsSecretMeta(lc),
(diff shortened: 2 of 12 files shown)


pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -0,0 +1,50 @@
...
+	return str
+}
+
+// GenerateObservabilityBundleConfigMap returns a configmap for
+// the observabilitybundle application to enable promtail.
+func GenerateLoggingCredentialsBasicSecret(lc loggedcluster.Interface) (v1.Secret, error) {
+
+	secret := v1.Secret{
+		ObjectMeta: LoggingCredentialsSecretMeta(lc),
+		Data:       map[string][]byte{},
```
TheoBrigitte, line 46: I think you should better use the `DataString` field here to write plain text data, otherwise you have to base64 encode it beforehand.

pkg/resource/logging-credentials/reconciler.go
```diff
@@ -0,0 +1,100 @@
...
+	if _, ok := currentSecret.Data["writeuser"]; !ok {
+		logger.Info("loggingCredentials - Adding writeuser key")
+		currentSecret.Data["writeuser"] = []byte("write")
+		secretUpdated = true
+	}
+	if _, ok := currentSecret.Data["writepassword"]; !ok {
+		logger.Info("loggingCredentials - Adding writepassword")
+		currentSecret.Data["writepassword"] = []byte(genPassword())
+		secretUpdated = true
+	}
```
TheoBrigitte, line 75: This should be moved to a function, so we can return `secretUpdated` instead of storing this inside this function.

pkg/resource/logging-credentials/reconciler.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// To be updated when secret is modified
+	var secretUpdated bool = false
+
+	err := r.Client.Get(ctx, types.NamespacedName{Name: secretName, Namespace: secretNamespace}, &currentSecret)
+	if err != nil {
+		if apimachineryerrors.IsNotFound(err) {
+			logger.Info("logging-credentials not found, initializing one")
+			// Create basic secret
+			currentSecret, err = GenerateLoggingCredentialsBasicSecret(lc)
```
TheoBrigitte, line 47: Here we are getting and generating secret with different name/namespace.
Please make sure we use the same values, when reading and when generating the secret.

TheoBrigitte (commented):
Looks overall good, password generation seems alright. We can always change this later if needed.

pkg/resource/logging-credentials/reconciler.go
hervenicol: I agree.
Is there such a thing as "package-global constants" that I could re-use across the package? Or should I write functions that will return the expected values?

pkg/resource/logging-credentials/logging_operator_secrets.go
hervenicol: I actually started by base64 encoding my strings, and they ended up being encoded twice. So I got rid of all the extra code and did this, and the created secret is "just enough" base64 encoded :grin: 
But DataString looks neat  :+1: , let me look if it can simplify my code

pkg/resource/logging-credentials/reconciler.go
hervenicol: Done, it looks much better like that indeed.

pkg/resource/logging-credentials/reconciler.go
hervenicol: Nevermind, should be global because other reconcilers will need it too.
So I'll do it as a config we can pass at startup.

pkg/resource/logging-credentials/logging_operator_secrets.go
TheoBrigitte: Ok fine, if this works as is.

pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	return secret, nil
+}
+
+// Update a LoggingCredentials secret if needed
+func UpdateLoggingCredentials(loggingCredentials v1.Secret) bool {
```
TheoBrigitte, line 53: Have you tested this ? I think you need to pass loggingCredentials as a pointer here if you want to update it in place.

pkg/resource/logging-credentials/reconciler.go
```diff
@@ -0,0 +1,79 @@
...
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+// ReconcileCreate ensures a secret exists for the given cluster.
+func (r *Reconciler) ReconcileCreate(ctx context.Context, lc loggedcluster.Interface) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	const secretNamespace = "monitoring"
+	const secretName = "logging-credentials"
```
TheoBrigitte, line 31: Usually its good to have those in a place where you can re-use them, as we later will need to refer to this object. That's why we usually have them in a ObjectMeta which is exposed vi an exported function.

pkg/resource/logging-credentials/reconciler.go
hervenicol: Done.

pkg/resource/logging-credentials/logging_operator_secrets.go
hervenicol: Short version: Yes, tested and it updates in place.

Longer version: I thought like you and did something like `newversion := UpdateLoggingCredentials(oldversion)` and took way longer than I should debugging why it didn't see differences between `newversion` and `oldversion` :rofl:

pkg/resource/logging-credentials/reconciler.go
hervenicol: Well, my latest commit created a common loggingConfig where we can put various config stuff, but maybe you have better ideas?

pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -0,0 +1,76 @@
...
+	secret := v1.Secret{
+		ObjectMeta: LoggingCredentialsSecretMeta(lc, config),
+		Data:       map[string][]byte{},
+	}
+
+	return secret, nil
+}
+
+// Update a LoggingCredentials secret if needed
+func UpdateLoggingCredentials(loggingCredentials v1.Secret) bool {
```
TheoBrigitte, line 54:
```suggestion
func UpdateLoggingCredentials(loggingCredentials *v1.Secret) bool {
```

pkg/resource/logging-credentials/reconciler.go
```diff
@@ -0,0 +1,78 @@
...
+			// Create basic secret
+			loggingCredentialsSecret, err = GenerateLoggingCredentialsBasicSecret(lc, r.LoggingConfig)
+		}
+		if err != nil {
+			return ctrl.Result{}, errors.WithStack(err)
+		}
+	}
+
+	// update the secret's contents if needed
+	secretUpdated := UpdateLoggingCredentials(loggingCredentialsSecret)
```
TheoBrigitte, line 50:
```suggestion
	secretUpdated := UpdateLoggingCredentials(&loggingCredentialsSecret)
```

pkg/resource/logging-credentials/logging_operator_secrets.go
hervenicol: Updated it to use a pointer.

pkg/resource/logging-credentials/reconciler.go
hervenicol: Changed it to retrieve info from `LoggingCredentialsSecretMeta(lc)`.
Yet I have some more tests to do because I currently have performance issues. :cry:

cmd/main.go
```diff
@@ -107,12 +108,18 @@ func main() {
 		Scheme: mgr.GetScheme(),
 	}
 
+	loggingSecrets := loggingcredentials.Reconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
```
fiunchinho, line 113: I think you are not using the `Scheme` field, so it could be removed from the reconciler.

cmd/main.go
```diff
@@ -107,12 +108,18 @@ func main() {
...
+	}
+
 	loggingReconciler := loggingreconciler.LoggingReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
 		Reconcilers: []reconciler.Interface{
 			&promtailReconciler,
 			&promtailWiring,
+			&loggingSecrets,
```
fiunchinho, line 122: What we do in other operators that we have is to run several reconcilers at the same time, instead of sequentially going one after the other. Have you consider using the other approach?

pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -0,0 +1,75 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// LoggingCredentialsSecretMeta returns metadata for the logging-operator credentials secret.
+func LoggingCredentialsSecretMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
+	return metav1.ObjectMeta{
+		Name:      "logging-credentials",
+		Namespace: "monitoring",
+	}
+}
```
fiunchinho, line 19: I think this could just be pair of constants for the namespace and name at the package level. Is there any advantage of having a function that always returns the same?

pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -0,0 +1,75 @@
...
+	str := b.String() // E.g. "ExcbsVQs"
+	return str
+}
+
+// GenerateObservabilityBundleConfigMap returns a configmap for
+// the observabilitybundle application to enable promtail.
+func GenerateLoggingCredentialsBasicSecret(lc loggedcluster.Interface) (v1.Secret, error) {
+
+	secret := v1.Secret{
+		ObjectMeta: LoggingCredentialsSecretMeta(lc),
```
fiunchinho, line 45: It would be cool to have some label with information about what component is managing the `Secret`, so that we can know if a secret was manually created or it was created by some automation.

pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -0,0 +1,75 @@
...
+		b.WriteRune(chars[rand.Intn(len(chars))])
+
+	}
+	str := b.String() // E.g. "ExcbsVQs"
+	return str
+}
+
+// GenerateObservabilityBundleConfigMap returns a configmap for
+// the observabilitybundle application to enable promtail.
+func GenerateLoggingCredentialsBasicSecret(lc loggedcluster.Interface) (v1.Secret, error) {
```
fiunchinho, line 42: The function returns a secret and an error. But I think most of the times where we need the secret, we actually need a pointer to a secret, so it may be more convenient to actually return it from here

```
GenerateLoggingCredentialsBasicSecret(lc loggedcluster.Interface) (*v1.Secret, error)
```

Thinking about this function, if we are never going to return an error, we don't need the second returned value and we could leave the signature as
```
GenerateLoggingCredentialsBasicSecret(lc loggedcluster.Interface) *v1.Secret
```

pkg/resource/logging-credentials/reconciler.go
```diff
@@ -0,0 +1,76 @@
...
+type Reconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+// ReconcileCreate ensures a secret exists for the given cluster.
+func (r *Reconciler) ReconcileCreate(ctx context.Context, lc loggedcluster.Interface) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	logger.Info(fmt.Sprintf("loggingcredentials checking secret %s/%s", LoggingCredentialsSecretMeta(lc).Namespace, LoggingCredentialsSecretMeta(lc).Name))
```
fiunchinho, line 30: We normally log using values instead of formatting strings. [Check this out](https://github.com/go-logr/logr#but-i-really-want-to-use-a-format-string).

pkg/resource/logging-credentials/reconciler.go
```diff
@@ -0,0 +1,76 @@
...
+	err := r.Client.Get(ctx, types.NamespacedName{Name: LoggingCredentialsSecretMeta(lc).Name, Namespace: LoggingCredentialsSecretMeta(lc).Namespace}, &loggingCredentialsSecret)
+	if err != nil {
+		if apimachineryerrors.IsNotFound(err) {
+			logger.Info("loggingcredentials secret not found, initializing one")
+			// Create basic secret
+			loggingCredentialsSecret, err = GenerateLoggingCredentialsBasicSecret(lc)
+		}
+		if err != nil {
+			return ctrl.Result{}, errors.WithStack(err)
+		}
```
fiunchinho, line 44: This is a strange flow. I don't know if it works by chance or deliberately but I think it's hard to understand at first glance that the `err` variable is getting overwritten by whatever error is returned by `GenerateLoggingCredentialsBasicSecret`. So the `if err != nil` from below that, it will check that error, instead of the one coming from fetching the secret from the k8s api. I would remove the error from `GenerateLoggingCredentialsBasicSecret`.

pkg/resource/logging-credentials/reconciler.go
```diff
@@ -0,0 +1,76 @@
...
+
+	return ctrl.Result{}, nil
+}
+
+// ReconcileDelete ensures a secret is removed for the current cluster
+func (r *Reconciler) ReconcileDelete(ctx context.Context, lc loggedcluster.Interface) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	logger.Info("loggingCredentials secret delete")
+
+	// Well, for the moment we don't have per-cluster creds, so we won't delete any.
```
fiunchinho, line 73: Don't we need to remove the `Secret` that was created during the reconciliation? Otherwise it will be left as a leftover.

pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -0,0 +1,75 @@
...
+	secret := v1.Secret{
+		ObjectMeta: LoggingCredentialsSecretMeta(lc),
+		Data:       map[string][]byte{},
+	}
+
+	return secret, nil
+}
+
+// Update a LoggingCredentials secret if needed
+func UpdateLoggingCredentials(loggingCredentials *v1.Secret) bool {
```
fiunchinho, line 53: `UpdateLoggingCredentials`  sounds like it's updating the secret when it gets outdated, but looking at the code, it won't update the contents of the fields once they are present. Is that the expected behavior?

TheoBrigitte (approved):
LGTM

pkg/resource/logging-credentials/reconciler.go
```diff
@@ -0,0 +1,76 @@
...
+			err = r.Client.Create(ctx, &loggingCredentialsSecret)
+		}
+		if err != nil {
+			return ctrl.Result{}, errors.WithStack(err)
+		}
+	} else {
+		logger.Info("loggingCredentials - up to date")
+	}
+
+	return ctrl.Result{}, nil
```
fiunchinho, line 65: I think the code would look simpler if you don't use the `else` code path, and just return early when you don't need to do anything else

```
	// commit our changes
	if !secretUpdated {
		return ctrl.Result{}, nil
	}
```

pkg/resource/logging-credentials/reconciler.go
```diff
@@ -0,0 +1,76 @@
...
+	// update the secret's contents if needed
+	secretUpdated := UpdateLoggingCredentials(&loggingCredentialsSecret)
+
+	// commit our changes
+	if secretUpdated {
+		logger.Info("loggingCredentials - Updating secret")
+		err = r.Client.Update(ctx, &loggingCredentialsSecret)
+		if apimachineryerrors.IsNotFound(err) {
+			logger.Info("loggingCredentials - Secret does not exist, creating it")
+			err = r.Client.Create(ctx, &loggingCredentialsSecret)
```
fiunchinho, line 56: As earlier, this is a odd way of doing error handling. Instead, I would explicitly check for the error
```
err = r.Client.Create(ctx, &loggingCredentialsSecret)
if err != nil {
}
```

fiunchinho (commented):
I have some suggestions. Also, I'm missing tests to make sure everything works as expected.

cmd/main.go
hervenicol: We considered it, and thought given there's not much cluster creation/deletion we would not benefit from tasks parallelization, and it would mix logs and thus make debugging harder.

However I'm thinking maybe we should have several controllers so that we watch different resources independently. Like:
* cluster is updated, we update loggingCredentials secret
* secret is updated, we update grafana datasource

Or could it be done at reconciler level?

pkg/resource/logging-credentials/logging_operator_secrets.go
hervenicol: Makes sense. Especially since I've already been tricked by using secret instead of pointer, and didn't understand why my original resource was updated! :sweat_smile:

cmd/main.go
hervenicol: Done.

cmd/main.go
TheoBrigitte: Do you maybe have an example on how this in done in another operator Jose ?

pkg/resource/logging-credentials/logging_operator_secrets.go
TheoBrigitte: So we could use this from other package later on.

pkg/resource/logging-credentials/reconciler.go
TheoBrigitte: Maybe in a future version we will move away from this. But for now it will stay.

cmd/main.go
```diff
@@ -110,7 +110,6 @@ func main() {
 
 	loggingSecrets := loggingcredentials.Reconciler{
 		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
```
TheoBrigitte: You are right, we do not need to pass Scheme along. We could simplify this.

pkg/resource/logging-credentials/logging_operator_secrets.go
hervenicol: Done in commit 81c1d49d6228b4aed1d0f5028ebe91129e6ed267

cmd/main.go
hervenicol: It has been done in commit 8d26017a1c852da239f0362797a5ec341467b1e3

pkg/resource/logging-credentials/logging_operator_secrets.go
hervenicol: Added with commit 476d19649ac56e906abe50833d8f4585e8aa6e2e

pkg/resource/logging-credentials/logging_operator_secrets.go
hervenicol: Yes, it's intended:
It will only add or remove credentials when clusters are added or deleted.
But today we have common credentials for all clusters, so it only checks these common credentials exists, or create them when missing.

pkg/resource/logging-credentials/reconciler.go
hervenicol: This code was removed / changed with commit 81c1d49d6228b4aed1d0f5028ebe91129e6ed267 :heavy_check_mark:

pkg/resource/logging-credentials/reconciler.go
hervenicol: Fixed in commit a4be7f073961511a9b3e9630e2d222817f2faf5c

pkg/resource/logging-credentials/reconciler.go
hervenicol: I agree 100%.
Done in commit 3abf188c947eb66d903101e9b256dc450f3c12e0

pkg/resource/logging-credentials/reconciler.go
hervenicol: Well, this one is a bit different:
- we try to update the secret
- if it fails because it does not exist (`IsNotFound`), we create it
- other errors are "real" errors and should fail
- If "create" fails, well it's a real failure too.

I updated this part of code in 3abf188c947eb66d903101e9b256dc450f3c12e0, maybe it makes more sense now?

hervenicol:
Ok, that was a lot of comments and suggestions, thanks @fiunchinho ! :heart: 
I'll merge now, but I keep working on this operator so feel free to add some comments, I may create a PR for fixes / improvements later.

cmd/main.go
fiunchinho: Yes, [we have two controllers here](https://github.com/giantswarm/aws-resolver-rules-operator/) and we are adding a third one in one of the open PRs.

pkg/resource/logging-credentials/logging_operator_secrets.go
fiunchinho: Can't you do that using constants?

pkg/resource/logging-credentials/reconciler.go
fiunchinho: If you add more code after the `if apimachineryerrors.IsNotFound(err) {}` block, you may inadvertently overwrite the error and forget to check the error returned in `err = r.Client.Create(ctx, loggingCredentialsSecret)`. That's why I propose to explicitly add the error handling for that
```
err = r.Client.Create(ctx, &loggingCredentialsSecret)
if err != nil {
}
```

PR #13, opened 2023-06-27
Helm chart

QuantumEnigmaa:
:warning: DO NOT MERGE :warning: 
Still need to be tested.

I'm especially not sure concerning the RBAC resource and if we still need to have a psp resource if we plan to release it only on the latest cluster versions.

=> Update Hervé 2023-07-07:
* Successfully deployed on `gauss`
* rbac fixed, but they could probably be more restrictive

So I think it's time to merge it!


helm/logging-operator/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+name: logging-operator
+description: The logging-operator manages loki and promtail related config.
+home: https://github.com/giantswarm/logging-operator
+version: [[ .Version ]]
```
QuentinBisson, line 5: Can you use abs instead of this?

helm/logging-operator/templates/_resource.tpl
```diff
@@ -0,0 +1,28 @@
...
+{{- define "resource.networkPolicy.name" -}}
+{{- include "resource.default.name" . -}}-network-policy
+{{- end -}}
+
+{{- define "resource.psp.name" -}}
+{{- include "resource.default.name" . -}}-psp
+{{- end -}}
+
+{{- define "resource.pullSecret.name" -}}
+{{- include "resource.default.name" . -}}-pull-secret
```
QuentinBisson, line 23: We do not need pull secret anymore

helm/logging-operator/templates/configmap.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "resource.default.name"  . }}
+  namespace: {{ include "resource.default.namespace"  . }}
+  labels:
+    {{- include "labels.common" . | nindent 4 }}
+data:
+  config.yml: |
+    server:
```
QuentinBisson, line 10: This config looks like an operator kit one :)

helm/logging-operator/templates/configmap.yaml
```diff
@@ -0,0 +1,25 @@
...
+      enable:
+        debug:
+          server: true
+      listen:
+        address: 'http://0.0.0.0:8000'
+    service:
+      alertmanager:
+        address: {{ .Values.alertmanagerAddress }}
```
QuentinBisson, line 18: I think alertmanager is not needed here :D

helm/logging-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+      volumes:
+      - name: {{ include "name" . }}-configmap
+        configMap:
+          name: {{ include "resource.default.name"  . }}
+          items:
+          - key: config.yml
+            path: config.yml
+      serviceAccountName: {{ include "resource.default.name"  . }}
+      securityContext:
+        runAsUser: {{ .Values.pod.user.id }}
```
QuentinBisson, line 32: I'd rather we use security context like upstream is doing rather that the old giantswarm way like https://github.com/prometheus-community/helm-charts/blob/44df643f30e8e623f4bb95f646de2662522eb554/charts/kube-prometheus-stack/values.yaml#L756

helm/logging-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+            {{- . | toYaml | nindent 10 }}
+          {{- end }}
+        resources:
+          requests:
+            cpu: {{ .Values.resources.requests.cpu }}
+            memory: {{ .Values.resources.requests.memory }}
+          limits:
+            cpu: {{ .Values.resources.limits.cpu }}
+            memory: {{ .Values.resources.limits.memory }}
+      imagePullSecrets:
```
QuentinBisson, line 67: No need for this anymore

helm/logging-operator/templates/network-policy.yaml
```diff
@@ -0,0 +1,20 @@
...
+metadata:
+  name: {{ include "resource.default.name"  . }}
+  namespace: {{ include "resource.default.namespace"  . }}
+  labels:
+    {{- include "labels.common" . | nindent 4 }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "labels.selector" . | nindent 6 }}
+  ingress:
```
QuentinBisson, line 12: You probably need a ciliumnetworkpolicy as well

helm/logging-operator/templates/pull-secret.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
QuentinBisson, line 1: Can be removed

helm/logging-operator/templates/_resource.tpl
QuantumEnigmaa: Great, removing it then :)

helm/logging-operator/templates/configmap.yaml
QuantumEnigmaa: True !

helm/logging-operator/Chart.yaml
QuantumEnigmaa: Then I'll have to put hard coded `version` and `appVersion` right ?

QuentinBisson: yes

helm/logging-operator/templates/network-policy.yaml
QuantumEnigmaa: Added it

helm/logging-operator/templates/configmap.yaml
QuantumEnigmaa: Should I change some things in it ?

helm/logging-operator/templates/cilium-network-policy.yaml
```diff
@@ -0,0 +1,16 @@
...
+  name: {{ include "resource.default.name"  . }}
+  namespace: {{ include "resource.default.namespace"  . }}
+spec:
+  endpointSelector:
+    matchLabels:
+      {{- include "labels.selector" . | nindent 6 }}
+  egress:
+    - toEntities:
+        - kube-apiserver
+{{ end }}
```
QuentinBisson, line 16:
```suggestion
{{ end }}

```

helm/logging-operator/templates/configmap.yaml
QuentinBisson: I'm not sure, let's get this running and change what needs to be changed :p

QuentinBisson (approved):

QuantumEnigmaa:
Not sure what is causing the build to fail : 
image

QuentinBisson:
Import of packages in the same project have to be split with outside packages and they must be imported last :(

QuantumEnigmaa:
>Import of packages in the same project have to be split with outside packages and they must be imported last :(

You mean instead of this : 
```go
import (
	"fmt"

	"github.com/giantswarm/logging-operator/pkg/key"
	"sigs.k8s.io/controller-runtime/pkg/client"
)
```
it should be this : 
```go
import (
	"fmt"

	"sigs.k8s.io/controller-runtime/pkg/client"

        "github.com/giantswarm/logging-operator/pkg/key"
)
```
?

QuantumEnigmaa:
Even after rearranging the imports the issue is still there  :/

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/giantswarm/logging-operator
 
-go 1.19
+go 1.18
```
hervenicol, line 3: why the rollback in go version?

helm/logging-operator/templates/configmap.yaml
hervenicol: True, logging-operator currently does not support config files. You can remove the whole volume and its references in `args`.
However, we'll have to set some `args` depending on some config values.

go.mod
hervenicol: Oh, because our build system does not support go 1.19! :fearful: 
image

Anyway fixing the build should not be in this "Helm chart" PR IMO.

hervenicol: Fixing in https://github.com/giantswarm/logging-operator/pull/16

hervenicol:
Woo! Build is fixed! :tada: (although I have no clue if the chart is OK. Let's work on this now).

hervenicol:
* Successfully deployed on `gauss`
* rbac fixed, but they could probably be more restrictive

So I think it's time to merge it!

PR #12, opened 2023-06-26
add grafana datasource

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/27146

Creates the DS for now, with read credentials from the logging-credentials.

Remark: maybe it should reconciliate the secret directly, rather than a cluster?

PR #14, opened 2023-06-27
create loki-multi-tenant auth

hervenicol:

Creates loki auth config, and trigger recycling of loki-multi-tenant-proxy pods once done.

Relies on #12 for `GetLogin` and `GetPass` functions.


QuentinBisson (approved):
I'm not happy about this tick tock workaround but let's move forward

PR #16, opened 2023-06-28
Fix go build

hervenicol:
`ci/circleci: go-build` was broken, let's fix it!

Build eventually fails when trying to publish the app, but that will be fixed when we have a helm chart!


QuantumEnigmaa:
That's nice ! 
Maybe we should add the change of go version from the mod file here as well ?

Dockerfile
```diff
@@ -1,32 +1,8 @@
...
-# by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go
-
-# Use distroless as minimal base image to package the manager binary
+# Use distroless as minimal base image to package the logging-operator binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
-COPY --from=builder /workspace/manager .
+ADD bin/logging-operator .
```
QuantumEnigmaa, line 5: Wouldn't it be better to keep a 2 stage Dockerfile with the first stage building the binary ? Otherwise does it means that the user will have to manually generate the binary with the makefile ?

hervenicol:
> That's nice ! Maybe we should add the change of go version from the mod file here as well ?

It's the first thing I did :thinking: 
If you see a missing go version downgrade, tell me where, because I may have missed it :sweat_smile:

Dockerfile
hervenicol: My point is:
- in the CI we've just built the binary before. Looks like a waste of resources to build it again.
- during dev we will probably mostly use the locally-built binary, not the docker image.
- if ever we want to use the image, we need to run `make` before to generate the binary, right.
- The Dockerfile seemed sketchy, trying to add only some specific go files. Looked like a recipe for broken Dockerfile to me, and some extra maintenance.

I agree in some cases it could be cool that the dockerfile builds the binary, but since the main build machinery is done with `make` it didn't make sense to me to have too many different build processes.

hervenicol: Already with architect and make doing different stuff, that's too many build processes for me :rofl:

QuantumEnigmaa: Ok that makes sense, especialy since the CI builds the binary

QuantumEnigmaa:
>If you see a missing go version downgrade, tell me where, because I may have missed it

No my bad, I missed thhe mod file x)

Dockerfile
hervenicol: I thought it made sense...
But as local `make` does stuff differently than circleci, either I keep adapting Makefile or I re-set Dockerfile to build the binary.
Any solution looks a bit ugly, I'm going for fixing Dockerfile :cry:

Makefile
```diff
@@ -40,7 +40,7 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=loggingoperator-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
QuentinBisson, line 43:
```suggestion
	$(CONTROLLER_GEN) rbac:roleName=logging-operator-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```

config/rbac/role.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   creationTimestamp: null
-  name: manager-role
+  name: loggingoperator-role
```
QuentinBisson, line 6:
```suggestion
  name: logging-operator-role
```

internal/controller/vintagemc_controller.go
```diff
@@ -80,7 +81,7 @@ func (r *VintageMCReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	loggedCluster := vintagemc.Object{
 		Object: service,
 	}
-	result, err = r.LoggingReconciler.Reconcile(ctx, loggedCluster)
+	_, err = r.LoggingReconciler.Reconcile(ctx, loggedCluster)
```
QuentinBisson, line 84: Why are we ignoring result here?

hervenicol: Was not used anyway, so it was creating errors at build.

QuentinBisson: Alright

QuentinBisson: Then apart from my 2 suggestions, I'm fine

QuentinBisson (approved):

QuantumEnigmaa (approved):

pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -24,22 +24,23 @@ func LoggingCredentialsSecretMeta(lc loggedcluster.Interface) metav1.ObjectMeta
...
 		"~=+%^*/()[]{}/!@#$?|")
 
-	var b strings.Builder
+	pass := make([]rune, length)
 	for i := 0; i < length; i++ {
-		b.WriteRune(chars[rand.Intn(len(chars))])
-
+		num, err := rand.Int(rand.Reader, big.NewInt(int64(len(chars))))
+		if err != nil {
+			return "", err
```
TheoBrigitte, line 39: We usually wrap errors for easier debugging when things go wrong
```suggestion
			return "", errors.WithStack(err)
```

internal/controller/vintagemc_controller.go
TheoBrigitte: this could return `ctrl.Result` to controller whether the resource should be re-queued or not. Currently we do not need that.

TheoBrigitte (approved):

PR #19, opened 2023-07-06
setup global logging disable flag

hervenicol:
New CLI flag to disable logging on all clusters.

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,6 +62,7 @@ func init() {
 
 func main() {
 	var enableLeaderElection bool
+	var disableLogging bool
 	var installationName string
 	var metricsAddr string
 	var probeAddr string
@@ -69,6 +70,7 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&disableLogging, "disable-logging", false, "disable logging for the whole installation")
 	flag.StringVar(&installationName, "installation-name", "unknown", "Name of the installation")
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -62,13 +62,15 @@ func init() {
 
 func main() {
 	var enableLeaderElection bool
+	var disableLogging bool
```
QuentinBisson, line 65: Why is that not enableLogging instead?

hervenicol: Because with the current libs/way args are managed, it seems a boolean is set to true when the parameter is passed as argument.

So it started from the CLI arg, and I wanted to keep it consistent across the whole app.

QuentinBisson: What do you prefer?

QuentinBisson: I would then see another flag like enable-vintage (false by default) that would be true only if we are in vintage clusters

hervenicol: I didn't manage to have a flag that is "true" by default and we can pass "=false" for it to be false.
I can do it with a string, but I thought that would be useless added complexity.

QuentinBisson: That's weird because it works on my machine :D 

Sample code:

```
  1 
  2 package main
  3 
  4 // Go provides a `flag` package supporting basic
  5 // command-line flag parsing. We'll use this package to
  6 // implement our example command-line program.
  7 import (
  8         "flag"
  9         "fmt"
 10 )
 11 
 12 func main() {
 13         // It's also possible to declare an option that uses an
 14         // existing var declared elsewhere in the program.
 15         // Note that we need to pass in a pointer to the flag
 16         // declaration function.
 17         var bvar bool
 18         flag.BoolVar(&bvar, "bvar", true, "a bool var")
 19 
 20         // Once all flags are declared, call `flag.Parse()`
 21         // to execute the command-line parsing.
 22         flag.Parse()
 23 
 24         fmt.Println("bvar:", bvar)
 25 }
```


go run main.go
bvar: true
> go run main.go  --bvar=false
bvar: false
> go run main.go
bvar: true

hervenicol: ok, yes it works if you do:
```
go run main.go --bvar=false
```

It won't if you do:
```
go run main.go --bvar false
```

I don't like that :disappointed: I think it's misleading.

hervenicol: ok, it will be managed in the helm chart anyway, fine with me! :grin: 
I updated the code.

main.go
```diff
@@ -131,8 +133,9 @@ func main() {
 		Client: mgr.GetClient(),
 	}
 
-	loggedcluster.O.LoggingEnabled = true
+	loggedcluster.O.EnableLoggingFlag = enableLogging
```
QuentinBisson, line 136: I would keep loggedcluster.O.LoggingEnabled = enableLogging and have logging enabled kept as a property everywhere but that's nice :)

QuentinBisson (approved):

PR #25, opened 2023-07-19
promtail-config reconciler: creates promtail-config as extra-values

hervenicol:

Promtail-config reconciler: creates promtail-config as extra-values


QuantumEnigmaa (approved):
LGTM if tested

PR #37, opened 2023-08-30
scrape logs from kube-system and giantswarm namespaces only

marieroque:
towards https://github.com/giantswarm/roadmap/issues/2758

### DIFF
--- a/pkg/resource/promtail-config/promtail-config.go
+++ b/pkg/resource/promtail-config/promtail-config.go
@@ -30,7 +30,8 @@ type promtail struct {
 }
 
 type promtailConfigSnippets struct {
-	ExtraScrapeConfigs string `yaml:"extraScrapeConfigs" json:"extraScrapeConfigs"`
+	ExtraRelabelConfigs string `yaml:"extraRelabelConfigs" json:"extraRelabelConfigs"`
+	ExtraScrapeConfigs  string `yaml:"extraScrapeConfigs" json:"extraScrapeConfigs"`
 }
 
 type promtailConfig struct {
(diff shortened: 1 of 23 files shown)


pkg/resource/promtail-config/promtail-config.go
```diff
@@ -75,6 +76,11 @@ func GeneratePromtailConfig(lc loggedcluster.Interface) (v1.ConfigMap, error) {
 			},
 			Config: promtailConfig{
 				Snippets: promtailConfigSnippets{
+					ExtraRelabelConfigs: `# we only scrape logs from kube-system and giantswarm namespaces
```
QuentinBisson, line 79: I think this should only apply to WCs though right?

marieroque: Hummm I totally forgot that point

pkg/resource/promtail-config/promtail-config.go
```diff
@@ -68,13 +70,24 @@ func ConfigMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
 // the promtail extra-config
 func GeneratePromtailConfig(lc loggedcluster.Interface) (v1.ConfigMap, error) {
 
+	// Scrape logs from kube-system and giantswarm namespaces only for WC clusters
+	extraRelabelConfigs := `[]`
+	if lc.GetObservabilityBundleConfigMap() == key.WCObservabilityBundleConfigMap {
```
hervenicol, line 75: As `WCObservabilityBundleConfigMap == CAPIObservabilityBundleConfigMap`, how does this make sure it does not apply to CAPI MCs?

QuentinBisson: Why should it not apply to CAPI MCs though?

QuentinBisson: oh nevermind

QuentinBisson: We need a isWorkloadCluster condition

QuentinBisson: In PMO it's mostly InstallationName == ClusterName

QuentinBisson (approved):
LGTM

QuentinBisson (approved):
LGTM

hervenicol:
:boom: 
Both build errors look like transient errors:
- gitleak, I've had the same on another repo at the same time. Now it builds.
- Aliyun, well, that's rather usual.

We should just retry build.

marieroque:
Tested on a `gaia` WC.

Only logs from `giantswarm` and `kube-system` namespaces:
image

PR #38, opened 2023-08-31
enable promtail deployment on WCs

QuantumEnigmaa:
Towards https://github.com/giantswarm/roadmap/issues/2761

Tested on a v19 WC and promtail is deployed but unable to send logs to loki on the MC. Need to solve this. 
Also gonna test it on a v18.4.2 WC.


internal/controller/vintagewc_controller.go
```diff
@@ -64,7 +64,7 @@ func (r *VintageWCReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, errors.WithStack(err)
 	}
 
-	logger.Info("Reconciling Vintage WV cluster", "name", cluster.GetName())
+	logger.Info("Reconciling Vintage WC cluster", "name", cluster.GetName())
```
QuentinBisson, line 67: 👍🏻

pkg/resource/promtail-wiring/reconciler.go
```diff
@@ -28,6 +28,8 @@ type Reconciler struct {
 // ReconcileCreate ensure user value configmap is set in observability bundle
 // for the given cluster.
 func (r *Reconciler) ReconcileCreate(ctx context.Context, lc loggedcluster.Interface) (ctrl.Result, error) {
+	isWC := lc.GetClusterName() != lc.GetInstallationName()
```
QuentinBisson, line 31: Can you check @marieroque PR and do the same?

pkg/resource/promtail-wiring/reconciler.go
```diff
@@ -42,16 +44,20 @@ func (r *Reconciler) ReconcileCreate(ctx context.Context, lc loggedcluster.Inter
...
 	}
 
-	// Set user value configmap in the app.
-	if setUserConfig(&currentApp, lc) {
-		logger.Info("promtailwiring updating")
-		// Update the app.
-		err := r.Client.Update(ctx, &currentApp)
-		if err != nil {
-			return ctrl.Result{}, errors.WithStack(err)
+	if !isWC {
```
QuentinBisson, line 47: I'd probably also check if this is vintage

pkg/vintagewc/cluster.go
```diff
@@ -46,5 +50,5 @@ func (o Object) GetObject() client.Object {
 
 // on vintage WC, use extraconfig
 func (o Object) GetObservabilityBundleConfigMap() string {
-	return "observability-bundle-logging-values"
+	return "observability-bundle-logging-extraconfig"
```
QuentinBisson, line 53: Is there a reason to remame it?

pkg/resource/promtail-toggle/observability_bundle_configmap.go
```diff
@@ -25,7 +25,7 @@ func ObservabilityBundleConfigMapMeta(lc loggedcluster.Interface) metav1.ObjectM
 		Namespace: lc.GetAppsNamespace(),
 		Labels: map[string]string{
 			// This label is used by cluster-operator to find extraconfig
-			"app.kubernetes.io/name": lc.AppConfigName("observability-bundle"),
+			"app.kubernetes.io/name": "observability-bundle",
```
QuentinBisson, line 28: That's true but not that is will only work on vintage WCs, Could you mention it in the comment?

pkg/vintagewc/cluster.go
QuantumEnigmaa: No particular reason, just gives directly the information about the nature of the cm

pkg/resource/promtail-toggle/observability_bundle_configmap.go
QuantumEnigmaa: Good point !

pkg/resource/promtail-wiring/reconciler.go
QuantumEnigmaa: Sure :)

pkg/vintagewc/cluster.go
QuentinBisson: But in that case that brings information about how it's mounted instead of what it is about :)

QuentinBisson (approved):
LGTM

pkg/resource/promtail-wiring/reconciler.go
QuantumEnigmaa: Done

pkg/vintagewc/cluster.go
QuantumEnigmaa: True but I guess it's less confusing for people to have really distinct name for those cms

QuentinBisson: Yep, but maybe we should name it with something related to promtail-toggle then but maybe we can discuss that later :)

pkg/resource/promtail-toggle/observability_bundle_configmap.go
hervenicol: This is why we were using a function from the cluster type-specific package, so we return a different value depending on whether we're on CAPI or vintage.

QuentinBisson: Oh then I guess this implementation needs to be fixed

QuantumEnigmaa:
Apparently CVEs were detected

QuantumEnigmaa:
This is kind of troublesome : even the latest stable version (0.28.1) of the `k8s.io/apiextensions-apiserver` still has the same CVE as the one with our current verion (0.26.1)

QuentinBisson:
Well if it does unless you want to fix it in apiextension-server upstream, I'd ignore the CVE with nancy, especially as we're not sure this will even work on 1.24 clusters

go.mod
```diff
@@ -4,74 +4,75 @@ go 1.18
...
+	github.com/onsi/ginkgo/v2 v2.11.0
+	github.com/onsi/gomega v1.27.8
 	github.com/pkg/errors v0.9.1
-	k8s.io/api v0.26.1
-	k8s.io/apimachinery v0.26.1
-	k8s.io/client-go v0.26.1
-	sigs.k8s.io/cluster-api v1.4.2
-	sigs.k8s.io/controller-runtime v0.14.5
+	k8s.io/api v0.29.0-alpha.0
+	k8s.io/apimachinery v0.29.0-alpha.0
```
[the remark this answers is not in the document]
QuantumEnigmaa: Is there a particular reason for it ?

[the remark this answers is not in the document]
QuentinBisson: kubernetes api is compatible up to 3 version and down to 2 versions. We run 1.24 so 1.29 would not be compatible with 1.24 ;)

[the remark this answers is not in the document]
QuantumEnigmaa: Oh ok sure !

pkg/resource/promtail-toggle/observability_bundle_configmap.go
hervenicol: I see this comment as "resolved", but yet we still have a hardcoded value that will not work in all situations.

QuentinBisson: Oh I was too eager Yesterday

QuantumEnigmaa: I am currently reworking on it

QuantumEnigmaa:
```
r.LoggingReconciler.Reconcile: capicluster.Object does not implement loggedcluster.Interface (missing ObservabilityBundleConfigLabelName method)
```
Same for vintagemc and vintagewc but I guess we just need to update the dependency

QuantumEnigmaa:
Tested and working :) 
I guess we're ready to merge

pkg/vintagewc/cluster.go
```diff
@@ -25,7 +25,15 @@ func (o Object) GetAppsNamespace() string {
...
-	return app
+	if app == "observability-bundle" {
+		return o.GetObject().GetName() + "-" + app
+	} else {
+		return app
+	}
+}
+
+func (o Object) ObservabilityBundleConfigLabelName(config string) string {
+	return config
```
QuentinBisson, line 36: Should this not have some kind of logic ?

QuantumEnigmaa: The method above (`AppConfigName`) was doing exactly the same before I updated it to match the needs for the WCs so I guess this is fine

QuentinBisson (approved):

PR #44, opened 2023-09-27
add unit tests

QuantumEnigmaa:
Towards https://github.com/giantswarm/giantswarm/issues/28196

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,11 @@ require (
 )
 
 require (
+	github.com/nxadm/tail v1.4.8 // indirect
+	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
+)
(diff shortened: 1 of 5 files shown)


internal/controller/vintagemc_controller_test.go
```diff
@@ -0,0 +1,148 @@
...
+				LeaderElectionID:       "5c8bbafe.x-k8s.io",
+			})
+			return err
+		}).Should(Succeed())
+
+		Expect(mgr).ToNot(BeEmpty())
+
+		promtailReconciler := promtailtoggle.Reconciler{
+			Client: mgr.GetClient(),
+			Scheme: mgr.GetScheme(),
```
QuentinBisson, line 97: And we probably do not even need a scheme here

PR #45, opened 2023-09-28
fix logging-operator promtail-wiring that deleted WC user configs

QuentinBisson:
Ensure we do not delete userconfigs for workload clusters to fix https://github.com/giantswarm/giantswarm/issues/28257


QuantumEnigmaa (approved):
LGTM

PR #48, opened 2023-10-02
Add audit logs

QuantumEnigmaa:
Towards https://github.com/giantswarm/giantswarm/issues/28346

This PR adds lines in the promtail config so that it can retrieve audit logs. Currently testing on a v19.1.0 cluster

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Audit logs in promtail config.
+
 ## [0.0.6] - 2023-09-28
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


QuentinBisson (approved):
LGTM, do you think we can deploy it only on control-plane nodes?

hervenicol:
> deploy it only on control-plane nodes?

Quite unlikely.
Promtail is a daemonset, all pods use the same config.

Maybe you have an idea how we should have a specific config for some nodes?

QuentinBisson:
@hervenicol I would think we could use the node role label for this but as long as it's not erroring out I'm fine

QuentinBisson:
@QuantumEnigmaa maybe it's because it's too long? Have you tried using a `|` ?

QuentinBisson:
See https://stackoverflow.com/a/21699210

QuantumEnigmaa:
Fixed and working on `gauss` : 
image

QuentinBisson (approved):

hervenicol (approved):

PR #53, opened 2023-10-14
Have per-cluster credentials and orgid.

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/28445

Having per-cluster orgid will help with performance analysis and provide per-cluster rate limiting.
On top of per-cluster orgid, this patch also sets per-cluster user. So we can enforce the org according to the user.

And a dashboard to show the difference:
image
* yellow: `giantswarm` tenant for all logs
* green: `gauss` tenant
* blue: my test WC's tenant

### DIFF
--- a/.nancy-ignore
+++ b/.nancy-ignore
@@ -1,12 +1,10 @@
-# pkg:golang/golang.org/x/net@v0.8.0
-CVE-2023-3978 until=2023-10-13
-
 # pkg:golang/google.golang.org/grpc@v1.52.0
-CVE-2023-32731 until=2023-10-13
+CVE-2023-32731 until=2024-01-02
 
 # pkg:golang/k8s.io/apiserver@v0.26.1
-CVE-2020-8561 until=2023-10-13
+CVE-2020-8561 until=2024-01-02
 
 # pkg:golang/golang.org/x/net@v0.8.0
-CVE-2023-3978 until=2023-10-13
+CVE-2023-3978 until=2024-01-02
+CVE-2023-39325 until=2024-01-02
 
--- a/pkg/resource/logging-credentials/logging_operator_secrets.go
+++ b/pkg/resource/logging-credentials/logging_operator_secrets.go
@@ -83,11 +83,12 @@ func GetPass(lc loggedcluster.Interface, credentialsSecret *v1.Secret, user stri
 	return string(pass), nil
 }
 
-// Update a LoggingCredentials secret if needed
-func UpdateLoggingCredentials(loggingCredentials *v1.Secret) bool {
+// AddLoggingCredentials - Add credentials to LoggingCredentials secret if needed
+func AddLoggingCredentials(lc loggedcluster.Interface, loggingCredentials *v1.Secret) bool {
 
 	var secretUpdated bool = false
 
+	// Always check credentials for "readuser"
 	if _, ok := loggingCredentials.Data["readuser"]; !ok {
 		loggingCredentials.Data["readuser"] = []byte("read")
 		secretUpdated = true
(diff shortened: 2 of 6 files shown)


QuentinBisson (approved):
LGTM

pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -100,18 +101,40 @@ func UpdateLoggingCredentials(loggingCredentials *v1.Secret) bool {
...
+}
+
+// RemoveLoggingCredentials - Remove credentials from LoggingCredentials secret
+func RemoveLoggingCredentials(lc loggedcluster.Interface, loggingCredentials *v1.Secret) bool {
+	var secretUpdated bool = false
+
+	// Check credentials for [clustername]
+	clusterName := lc.GetClusterName()
+
+	if _, ok := loggingCredentials.Data[fmt.Sprintf("%suser", clusterName)]; ok {
```
QuentinBisson, line 130: Maybe extract the 2 sprintf into variables for performance?

pkg/resource/loki-auth/loki-auth.go
```diff
@@ -46,46 +48,74 @@ func LokiAuthSecretMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
...
 }
 
+// listWriteUsers returns a map of users found in a credentialsSecret
+func listWriteUsers(credentialsSecret *v1.Secret) []string {
+	var usersList []string
+	for myUser := range credentialsSecret.Data {
+
+		userTrimmed := strings.TrimSuffix(myUser, "user")
+		// bypass entries that are not a user
+		if userTrimmed == myUser {
```
QuentinBisson, line 58: I'd use this instead to remove the continue

```suggestion
		if userTrimmed != myUser && userTrimmed != "read" {
		  usersList = append(usersList, userTrimmed)
		}
```

pkg/resource/loki-auth/loki-auth.go
```diff
@@ -46,46 +48,74 @@ func LokiAuthSecretMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
...
-	readUser, err := loggingcredentials.GetLogin(lc, credentialsSecret, "read")
-	if err != nil {
-		return v1.Secret{}, errors.WithStack(err)
+	// Init empty users structure
+	values := Values{
+		Users: []user{},
 	}
+	// Prepare read user's orgid with default values
+	// make sure to have at least 2 tenants for read user, to prevent writing with this user
+	readOrgid := "giantswarm|default"
```
QuentinBisson, line 82: Maybe make those a constant like DefaultReadOrgIDs?

pkg/resource/logging-credentials/logging_operator_secrets.go
hervenicol: I don't think there's much of a performance impact because it does not seem to be run often.
However I agree that's quite ugly.

But I was more thinking about changing the way this secret is built, but in a separate/future PR.
Today:
```
data:
  gausspassword: [base64 pass]
  gaussuser: Z2F1c3M=
  readpassword: [base64 pass]
  readuser: cmVhZA==
  sm2t7password: [base64 pass]
  sm2t7user: c20ydDc=
```

I think we'd better have:
```
data:
  gauss: [base64 user and pass structure]
  read: [base64 user and pass structure]
  sm2t7: [base64 user and pass structure]
```

pkg/resource/loki-auth/loki-auth.go
hervenicol: My mind understands better the logic of:
* check for all known exceptions
* then do the job

Also it reduces the number of code imbrication levels.
For these two reasons, I did it this way to make it easier to read.

But that's quite subjective, and you're a more experienced developer than me so I can do it as you suggested.

pkg/resource/logging-credentials/logging_operator_secrets.go
QuentinBisson: That New format would bé better. The reason I was proposing a function is that this specific printf is repeated often and i'm a huge believer of DRY

pkg/resource/loki-auth/loki-auth.go
QuentinBisson: Whether you use continue or not does not really change much but i'd probably have only 1 if as it's not really nested but i'm definitely not blocking

.nancy-ignore
```diff
@@ -1,12 +1,10 @@
...
-
 # pkg:golang/google.golang.org/grpc@v1.52.0
-CVE-2023-32731 until=2023-10-13
+CVE-2023-32731 until=2024-01-02
 
 # pkg:golang/k8s.io/apiserver@v0.26.1
-CVE-2020-8561 until=2023-10-13
+CVE-2020-8561 until=2024-01-02
 
 # pkg:golang/golang.org/x/net@v0.8.0
```
QuentinBisson, line 7: Should we maybe upgrade to v0.17.0?

QuentinBisson (approved):

PR #54, opened 2023-10-16
change logging-credentials secret format

hervenicol:
Before:
```
data:
  gausspassword: [base64 pass]
  gaussuser: Z2F1c3M=
  readpassword: [base64 pass]
  readuser: cmVhZA==
  sm2t7password: [base64 pass]
  sm2t7user: c20ydDc=
```

After:
```
data:
  gauss: [base64 pass structure]
  read: [base64 pass structure]
  sm2t7: [base64 pass structure]
```

The base64 pass structure is simple (but can be expanded in the future if needed):
```
password: [password]
```

### DIFF
--- a/pkg/resource/logging-credentials/logging_operator_secrets.go
+++ b/pkg/resource/logging-credentials/logging_operator_secrets.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -20,6 +21,10 @@ const (
 	LoggingCredentialsName      = "logging-credentials"
 	LoggingCredentialsNamespace = "monitoring"
 )
+
+type user struct {
+	Password string `yaml:"password" json:"password"`
+}
 
 // LoggingCredentialsSecretMeta returns metadata for the logging-operator credentials secret.
 func LoggingCredentialsSecretMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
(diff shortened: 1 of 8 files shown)


QuentinBisson:
Does that mean we have a base64 of a base64?

pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -21,6 +22,10 @@ const (
 	LoggingCredentialsNamespace = "monitoring"
 )
 
+type user struct {
```
QuentinBisson, line 25: Should that struct be named credentials or userCredentials?

pkg/resource/logging-credentials/logging_operator_secrets.go
```diff
@@ -64,60 +69,71 @@ func GenerateLoggingCredentialsBasicSecret(lc loggedcluster.Interface) *v1.Secre
 	return &secret
 }
 
-func GetLogin(lc loggedcluster.Interface, credentialsSecret *v1.Secret, user string) (string, error) {
-
-	login, ok := credentialsSecret.Data[fmt.Sprintf("%suser", user)]
+func GetPass(lc loggedcluster.Interface, credentialsSecret *v1.Secret, username string) (string, error) {
```
QuentinBisson, line 72: Maybe call it GetPassword :)

QuentinBisson:
I like this format way better :)

I'm wondering if we should separate read and have a tenants subsection but that is not that important

hervenicol:
> Does that mean we have a base64 of a base64?

That would not be uncommon, but no.
- the secret is b64-encoded, as any secret.
- the password inside the secret is not encoded.

pkg/resource/logging-credentials/logging_operator_secrets.go
hervenicol: Done.

pkg/resource/logging-credentials/logging_operator_secrets.go
hervenicol: Done.

hervenicol:
> I like this format way better :)

Thanks :heart: 

> I'm wondering if we should separate read and have a tenants subsection but that is not that important

I'm not sure yet what's happening to `read` in the future, when we have multi-tenancy on the read path. I'm happy if we don't touch it for now :grin:

QuentinBisson (approved):
Lgtm

PR #62, opened 2023-10-18
Add capi support

QuentinBisson:
This PR adds CAPI support to logging operator


pkg/resource/promtail-wiring/reconciler.go
```diff
@@ -67,25 +82,39 @@ func (r *Reconciler) ReconcileDelete(ctx context.Context, lc loggedcluster.Inter
...
+			logger.Info("promtailwiring - app not found")
+			return ctrl.Result{}, nil
 		}
+		return ctrl.Result{}, errors.WithStack(err)
+	}
 
-		// Unset user value configmap in the app.
-		if unsetUserConfig(&currentApp, lc) {
+	if lc.IsVintage() {
+		if common.IsWorkloadCluster(lc) {
```
hervenicol, line 100: I was really hoping we could avoid these conditions by having interfaces :cry: but I don't have a better suggestion at the moment.

hervenicol:
You have to differenciate MC from WCs, even on Capi, because the list of logs to retrieve differs.

I was wondering if that could be externalized in a config, so we can select which logs to retrieve for each clusters.
By default, we would have:
* only select logs from kube-system and giantswarm namespaces
* a custom configmap deployed via flux on all MCs (collections?) to retrieve all logs.

Not sure it's part of this PR though :sweat_smile:

QuentinBisson:
The différentiation is actually done already in the promtail-config resource as we're using common.IsWorkloadCluster there :)

pkg/resource/promtail-wiring/reconciler.go
QuentinBisson: We could do with interfaces but we'd need to have this setExtraConfig in the interface implémentation :)

hervenicol: Well, I'd prefer, as that's supposed to be their role.

QuentinBisson:
@hervenicol this actually looks way better, but that means I have to test it everywhere :D

pkg/common/common.go
```diff
@@ -8,13 +8,11 @@ import (
...
 func IsLoggingEnabled(lc loggedcluster.Interface) bool {
-
 	// Logging should be enabled when all conditions are met:
 	//   - logging label is set and true on the cluster
 	//   - cluster is not being deleted
-	//   - TODO(theo) global logging flag is enabled
-
-	return lc.GetLoggingLabel() == "true" && lc.GetDeletionTimestamp().IsZero() && lc.GetEnableLoggingFlag()
+	//   - global logging flag is enabled
+	return lc.IsLoggingEnabled() && lc.GetDeletionTimestamp().IsZero() && lc.GetEnableLoggingFlag()
```
hervenicol, line 15: We had named the lc function `GetLoggingLabel` because it is not enough of a condition for logging to be enabled.
As seen here, `common.IsLoggingEnabled` depends on multiple conditions, and having the logging label is one of them.

I thing both function namings are confusing, but I obviously prefer `GetLoggingLabel` because that's what I chose when I wrote it, after hesitating between those 2 names :grin:

QuentinBisson: Well, I would agree but if you look at the function code, it now does more than just getting a label value right? Ir actually checks that logging is enabled for the logged cluster

QuentinBisson: We could rename it to hasLoggingEnabled if you prefer?

hervenicol: ok for `hasLoggingEnabled`, at least that's a different function name.

QuentinBisson: Done :)

QuentinBisson:
Dropping in favor of https://github.com/giantswarm/logging-operator/pull/81 and https://github.com/giantswarm/object-storage-operator/pull/26

PR #76, opened 2023-10-31
Push to capa

QuentinBisson:
Towards https://github.com/giantswarm/roadmap/issues/2937

Let's deploy the logging operator on capa so we can use opsctl deploy :)


TheoBrigitte (approved):
LGTM

PR #79, opened 2023-11-02
prepare capi support

QuentinBisson:
This PR sets some things so the CAPI support PR is easier to digest


helm/logging-operator/templates/deployment.yaml
```diff
@@ -37,7 +37,10 @@ spec:
         {{- else -}}
           - -enable-logging=false
         {{- end }}
-        - -installation-name={{ .Values.loggingOperator.installationName }}
+        - -installation-name={{ .Values.managementCluster.name }}
+        - -installation-provider={{ .Values.managementCluster.provider.kind }}
+        - -installation-region={{ .Values.aws.region }}
+        - -installation-base-domain={{ .Values.managementCluster.baseDomain }}
```
QuentinBisson, line 43: This is needed for IRSA

helm/logging-operator/templates/deployment.yaml
```diff
@@ -37,7 +37,10 @@ spec:
         {{- else -}}
           - -enable-logging=false
         {{- end }}
-        - -installation-name={{ .Values.loggingOperator.installationName }}
+        - -installation-name={{ .Values.managementCluster.name }}
+        - -installation-provider={{ .Values.managementCluster.provider.kind }}
+        - -installation-region={{ .Values.aws.region }}
```
QuentinBisson, line 42: THis is needed to create an IAM client

helm/logging-operator/templates/deployment.yaml
```diff
@@ -37,7 +37,10 @@ spec:
         {{- else -}}
           - -enable-logging=false
         {{- end }}
-        - -installation-name={{ .Values.loggingOperator.installationName }}
+        - -installation-name={{ .Values.managementCluster.name }}
+        - -installation-provider={{ .Values.managementCluster.provider.kind }}
```
QuentinBisson, line 41: Needed to be able to create the loki role on capa MCs only

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/giantswarm/logging-operator
 
-go 1.18
+go 1.20
```
QuentinBisson, line 3: Bumps dependencies

helm/logging-operator/values.yaml
```diff
@@ -2,6 +2,11 @@ global:
   podSecurityStandards:
     enforced: false
 
+aws:
+  accessKeyID: accesskey
```
QuentinBisson, line 6: Needed to be able to connect to the aws api

helm/logging-operator/templates/deployment.yaml
```diff
@@ -37,7 +37,10 @@ spec:
         {{- else -}}
           - -enable-logging=false
         {{- end }}
-        - -installation-name={{ .Values.loggingOperator.installationName }}
+        - -installation-name={{ .Values.managementCluster.name }}
```
hervenicol, line 40: This requires a change in `config` repo.

QuentinBisson: Yep it's done

QuentinBisson: Let me track it After lunch

QuentinBisson: Sée https://github.com/giantswarm/config/blob/b95dba75f82525f9ece4141b4eb312d4f3c4902e/default/apps/logging-operator/configmap-values.yaml.template#L2

helm/logging-operator/templates/deployment.yaml
hervenicol: "region" is probably AWS-specific.
Maybe that should be made clear in its name?

QuentinBisson: Well it is not really as it's also set in azure but I need to move where it is defined in config

pkg/capicluster/cluster.go
```diff
@@ -14,12 +18,24 @@ type Object struct {
...
-	return value
+	loggingLabelValue, ok := labels[key.LoggingLabel]
+	if !ok {
+		return true
+	}
+
+	loggingEnabled, err := strconv.ParseBool(loggingLabelValue)
+	if err != nil {
+		return false
+	}
```
hervenicol, line 37: If there is no label, it returns `true` => this means logging becomes enabled by default, right?
But if there's an error in `ParseBool`, it returns `false` => why not sticking to the default behaviour (`true`) then?

QuentinBisson: Makes sense

QuentinBisson: Fixed

hervenicol (approved):
I'm not very happy that it has to access AWS api, but apart from that LGTM.

PR #91, opened 2023-11-16
Move logged cluster types under pkg/logged-cluster so it's easier to …

QuentinBisson:
Small refactoring so it's easier to change the logged-cluster api (adding or removing a field is done in the same package)


QuantumEnigmaa (approved):
LGTM

PR #94, opened 2023-11-24
Remove system audit logs

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/28842

This PR does:

* Drop system audit logs from Promtail's scrape targets
The amount of system audit logs is order of magnitude greater than other system logs.
This is causing a high load on Loki. We can always add those back later on.

* Fix empty systemd_unit label
The systemd_unit label was empty in certain cases (e.g. kernel logs), this was fixed by adding the syslog identifer as systemd_unit label.

---

### Investigation
1. Identified `system-logs` and `kubernetes-pods` as having most log lines

Screenshot_2023-11-24_17-32-25

2. Checking top log lines per pods: `nginx-ingress-controller` and `coredns` are outliers, but still we need them

image

3. Significant change after dropping `system-logs`/`audit` which produces 20'000 log lines while other system-logs are at ~200.

image

4. `systemd_unit` label missing on some system logs
Screenshot_2023-11-24_17-16-29

5. After `systemd_unit` label fix
Screenshot_2023-11-24_17-14-12


QuentinBisson:
I'm not sure I agréé on dropping those logs. WE need systèm audit logs for sécurity reasons

hervenicol:
> I'm not sure I agréé on dropping those logs. WE need systèm audit logs for sécurity reasons

I agree with you: we need to keep audit logs for security.
But I'm not sure Loki is the place to store them. At least for now.
And if we want to store them in Loki, they probably deserve a better `scrape_job` than `system_logs`, with some specific cleanup and better labels.

So it seemed fine to us to just remove them for now.

TheoBrigitte:
@QuentinBisson I updated my first message and created a follow-up task to add those back later https://github.com/giantswarm/giantswarm/issues/29095

Can we proceed ?

QuentinBisson (approved):

QuentinBisson:
@TheoBrigitte should we go ahead ?

PR #103, opened 2023-12-13
Fix podmonitor

hervenicol:
Towards https://github.com/giantswarm/giantswarm/issues/29262

With this patch Prometheus-agent scrapes data from the pod.

### DIFF
--- a/helm/logging-operator/templates/deployment.yaml
+++ b/helm/logging-operator/templates/deployment.yaml
@@ -45,6 +45,10 @@ spec:
           {{- with .Values.securityContext.container }}
             {{- . | toYaml | nindent 10 }}
           {{- end }}
+        ports:
+        - containerPort: 8080
+          name: http
+          protocol: TCP
         resources:
           requests:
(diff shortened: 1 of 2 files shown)


helm/logging-operator/templates/deployment.yaml
```diff
@@ -45,6 +45,10 @@ spec:
           {{- with .Values.securityContext.container }}
             {{- . | toYaml | nindent 10 }}
           {{- end }}
```
QuentinBisson, line 47: Do we need that for object-storage operator as well ?

QuentinBisson (approved):

helm/logging-operator/templates/deployment.yaml
hervenicol: I couldn't say for sure, that's not part of this PR. But that looks like a good idea yes.

QuentinBisson: Considering I copied this pod monitor I think so 😅

PR #155, opened 2024-06-12
Fix reconciliation errors

marieroque:
towards https://github.com/giantswarm/giantswarm/issues/31039

That PR fixes the reconciliation errors when we add or remove the finalizer on the Cluster CR if it has been updated in the meantime.


pkg/logging-reconciler/logging_reconciler.go
```diff
@@ -42,9 +44,14 @@ func (l *LoggingReconciler) reconcileCreate(ctx context.Context, lc loggedcluste
 		logger.Info("adding finalizer", "finalizer", key.Finalizer)
 		controllerutil.AddFinalizer(lc, key.Finalizer)
 		err := l.Client.Update(ctx, lc.GetObject())
-		if err != nil {
-			logger.Error(err, "failed to add finalizer to logger cluster", "finalizer", key.Finalizer)
-			return ctrl.Result{}, errors.WithStack(err)
+		// We check if the error is a conflict error, which means that the object has been updated since we last fetched it.
+		if apimachineryerrors.IsConflict(err) {
+			return ctrl.Result{RequeueAfter: time.Duration(5 * time.Minute)}, nil
```
QuentinBisson, line 49: We probably want to requeue faster though

marieroque: How many time do you suggest ? 1m ?

QuentinBisson (approved):

marieroque:
We try an other solution with adding a rate limiting on the CAPIController to limit the number of errors.
Those errors are due to a big number of changes on the ClusterCR (all operators adding their finalizer).
All thoses changes are executing at the same time (exact second)...

internal/controller/capicluster_controller.go
```diff
@@ -97,3 +104,14 @@ func (r *CapiClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		).
 		Complete(r)
 }
+
+// ControllerRateLimiter returns a rate limiter to control the rate of reconciliation.
+// Taken from https://github.com/kubernetes/client-go/blob/20732a1bc198ab57de644af498fa75e73fa44c08/util/workqueue/default_rate_limiters.go#L39-L45
```
QuentinBisson, line 109: Can you explain why you add this rate limiter?

pkg/logging-reconciler/logging_reconciler.go
```diff
@@ -40,12 +40,16 @@ func (l *LoggingReconciler) reconcileCreate(ctx context.Context, lc loggedcluste
 
 	if !controllerutil.ContainsFinalizer(lc, key.Finalizer) {
 		logger.Info("adding finalizer", "finalizer", key.Finalizer)
+
 		controllerutil.AddFinalizer(lc, key.Finalizer)
-		err := l.Client.Patch(ctx, lc.GetObject())
+		// We choose to make a Patch instead of an Update to avoid error when the object is updated by another operator.
```
QuentinBisson, line 45: Maybe we want to do that in the o11yop as well?

internal/controller/capicluster_controller.go
TheoBrigitte: We though the rate limiter would help reduce the amount of incoming request and therefore reduce the amount of errors, but it doesn't as it only act on internal retries. So we abandoned this idea.

pkg/logging-reconciler/logging_reconciler.go
TheoBrigitte: Yes this seems to solve the issue, so we should port this to observability operator. I would suggest first we release this and wait to see if it works as expected.

TheoBrigitte (approved):
Works for me.

Just a reminder to add some comments.

pkg/logging-reconciler/logging_reconciler.go
marieroque: Yes... same errors on that operator

marieroque:
✅ Tested on `grizzly`
✅ Tested on `gaia`

PR #158, opened 2024-06-26
enhancement: change grafana agent from daemonset to deployment

QuentinBisson:
Closes https://github.com/giantswarm/giantswarm/issues/30317

This PR changes the configuration of the grafana agent used to collect kubernetes events in clusters from a daemonset to a deployment. As the grafana agent is only used to collect events, there is no reason to deploy it as a daemonset.

This should reduce the resource usage of our platform accross all clusters once the logging operator is released with this change


QuentinBisson:
Tested on grizzly:
image

We are still receiving events:
image

TheoBrigitte (commented):
Do we just go from deamonset to 1 deployment? Is this enough ?

QuentinBisson:
It should be yes as they watch the same events

hervenicol (approved):
The `GrafanaAgent` definition is quite hard to read, with the mix of structured data and free text in the `Content` part :cry: 

But otherwise LGTM

hervenicol:
> Do we just go from deamonset to 1 deployment? Is this enough ?

I guess the implied question was: that's only 1 pod, so there is no more redundancy. What happens when it breaks, it there a risk of missing some events?

QuentinBisson:
I think the agent should bé able to get évents as long as it's up. Do you prefer me to configure 2-3 replicas?

TheoBrigitte:
> I think the agent should bé able to get évents as long as it's up. Do you prefer me to configure 2-3 replicas?

Yes I would prefer having 3 replicas.

QuentinBisson:
Actually, maybe we should make it a single replica statefulset ? https://grafana.com/docs/agent/latest/flow/reference/components/loki.source.kubernetes_events/#component-behavior

hervenicol:
Well, yes, it looks like a single-replica statefulset may be the way to go. With a tiny pv.

QuentinBisson:
Moved to a single replicas statefulset with a 1 Gi pv

pkg/resource/grafana-agent-config/grafana-agent-config.go
```diff
@@ -93,6 +105,10 @@ loki.write "default" {
 }`,
 				},
 			},
+			Controller: controller{
+				Replicas: 1,
+				Type:     deployment,
```
marieroque, line 110: Don't we decide to use sts ?

QuentinBisson: We decided not to on friday when you cut the call :D

marieroque: Can you add a comment explaining why we decided to use a deployment with 1 replica here or in the linked ticket?

QuentinBisson: Sure, we decided to use a deployment because statefulsets should be used for storing metrics (when using a wal) and we are fine with losing a few events when the pod restarts. We're mostly concerned about creating a PV on all WCs

marieroque (approved):

PR #164, opened 2024-07-15
add `--logging-agent` flag to toggle betwen Promtail and Alloy as logging agent

TheoBrigitte:

This PR adds `--logging-agent` flag which can take 2 distinct values: `promtail` and `alloy`. The flag is used to toggle between Promtail and Alloy as the logging agent for the cluster.

- Move Promtail secret generation from logging-secret.go > promtail-logging-secret.go
- Rename GenerateLoggingSecret > GeneratePromtailLoggingSecret
- Add alloy-logging-secret.go
- Add Alloy values and config as template files
- Move backOffMaxRetries and loki url format into common
- Add logic to toggle between Promtail and Alloy; delete one wen the other is enabled

The provided logging.alloy.template was generated with `alloy convert --source-format=promtail --output logging.alloy promtail.yaml`.

That's a long PR, I can split this into smaller chunk if needed.


TheoBrigitte:
Tested successfully on `golem` :heavy_check_mark: 

* log volume is same as with Promtail (besides the spike on restart at 16:30)

image


* node name is there for audit logs

image

* audit logs are correctly filtered out from systemd logs

image

pkg/resource/logging-secret/reconciler.go
```diff
@@ -44,7 +44,17 @@ func (r *Reconciler) ReconcileCreate(ctx context.Context, lc loggedcluster.Inter
 	}
 
 	// Get desired secret
-	desiredLoggingSecret, err := GenerateLoggingSecret(lc, &loggingCredentialsSecret, lokiURL)
+	// Get desired config
+	var desiredLoggingSecret v1.Secret
+	switch lc.GetLoggingAgent() {
+	case "promtail":
+		desiredLoggingSecret, err = GeneratePromtailLoggingSecret(lc, &loggingCredentialsSecret, lokiURL)
```
QuentinBisson, line 51: I think we should add an interface that generates the logging secret and the deciding code will be there and not in the reconciler because this is missing the app version (if the observability bundle does not support alloy then that's not working at all)

pkg/resource/logging-agents-toggle/observability_bundle_configmap.go
```diff
@@ -27,9 +27,26 @@ func GenerateObservabilityBundleConfigMap(lc loggedcluster.Interface, observabil
 	if observabilityBundleVersion.LT(semver.MustParse("1.0.0")) {
 		promtailAppName = "promtail-app"
 	}
-	appsToEnable[promtailAppName] = app{
-		Enabled: true,
+
+	switch lc.GetLoggingAgent() {
+	case "promtail":
```
QuentinBisson, line 32: This is missing a check if the observability bundle even has alloy

pkg/resource/logging-secret/alloy/logging-secret.yaml.template
```diff
@@ -0,0 +1,36 @@
+alloy:
```
QuentinBisson, line 1: 1. Do we really need to override all of this just to set some config?
2. Why do we have to use templates? This reminds me of the PMO stuff we have and I'm not sure I like it :( 
3. Could you add comments explaining what the template do?

pkg/resource/logging-secret/alloy/logging.alloy.template
```diff
@@ -0,0 +1,265 @@
+discovery.kubernetes "kubernetes_pods" {
```
QuentinBisson, line 1: I'm pretty sure we will have to change this but this is a translation of the current promtail config

pkg/resource/logging-agents-toggle/observability_bundle_configmap.go
QuentinBisson: i would rather we default to promtail if it's not there (checking olly bundle version)

pkg/resource/logging-secret/reconciler.go
TheoBrigitte: Can you elaborate on what this interface would look like ?

pkg/resource/logging-agents-toggle/observability_bundle_configmap.go
TheoBrigitte: something like `observabilityBundleVersion.GE(semver.MustParse("1.4.0")` ?

pkg/resource/logging-secret/alloy/logging-secret.yaml.template
TheoBrigitte: 1. Yes we need those override to make alloy work as a logging agent
2. Using template save the complexity of defining all the yaml structure as Go struct, its then easier to update those values later on.
3. I'll add that

pkg/resource/logging-secret/alloy/logging.alloy.template
TheoBrigitte: Yes this config file is a translation of the current promtail config.

pkg/resource/logging-agents-toggle/observability_bundle_configmap.go
QuentinBisson: Yes that would be nice :)

pkg/resource/logging-secret/alloy-logging-secret.go
```diff
@@ -0,0 +1,106 @@
...
+		return v1.Secret{}, err
+	}
+
+	return secret, nil
+}
+
+func generateAlloyLoggingConfig(lc loggedcluster.Interface, credentialsSecret *v1.Secret, lokiURL string) (v1.Secret, error) {
+	var values bytes.Buffer
+
+	t, err := template.New("logging-config.yaml").Funcs(sprig.FuncMap()).Parse(alloyLoggingSecret)
```
QuentinBisson, line 36: I think templates should be parsed in a constructor function to avoid doing it on every run because this is using some resources

pkg/resource/logging-secret/alloy-logging-secret.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+//go:embed alloy/logging.alloy.template
+var alloyLogging string
+
+//go:embed alloy/logging-secret.yaml.template
+var alloyLoggingSecret string
+
+func GenerateAlloyLoggingSecret(lc loggedcluster.Interface, credentialsSecret *v1.Secret, lokiURL string) (v1.Secret, error) {
+	secret, err := generateAlloyLoggingConfig(lc, credentialsSecret, lokiURL)
```
QuentinBisson, line 25: I'm a bit confused that we use a function named generateAlloyLoggingConfig in a function named generateAlloyLoggingSecret. Do we need that second function at all?

pkg/resource/logging-secret/alloy-logging-secret.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/Masterminds/sprig/v3"
+
+	"github.com/giantswarm/logging-operator/pkg/common"
+	loggedcluster "github.com/giantswarm/logging-operator/pkg/logged-cluster"
+	loggingcredentials "github.com/giantswarm/logging-operator/pkg/resource/logging-credentials"
+)
+
+//go:embed alloy/logging.alloy.template
+var alloyLogging string
```
QuentinBisson, line 19: Can we not group them using var (
  alloyLogging string
  alloyLoggingSecret string
)

pkg/resource/logging-secret/alloy/logging-secret.yaml.template
```diff
@@ -0,0 +1,45 @@
...
+# - Running as root user is required in order to be able to read log files within
+#   /var/log/journal and /run/log/journal directories.
+# - NODENAME env var is used as additional label for kubernetes_audit logs.
+alloy-logs:
+  alloy:
+    configMap:
+      create: true
+      content: |-
+{{ .AlloyConfig | indent 8 }}
+    extraEnv:
```
QuentinBisson, line 16: I feel like this as well as other properties should be part of the app itself and not this

pkg/resource/logging-secret/reconciler.go
QuentinBisson: I think i would have kept the logging-secret.go resource and have the logging agent flag done there. 

Like that you can share the secret creation and only the content changes?

QuentinBisson:
Regarding the overall PR, do we want to:
- Migrate grafana agent as well as the translation should be easy?
- Migrate old promtail and grafana-agent configs to a template?
- Try out default alloy pod discovery with structured metadata instead of the cumbersome config?

hervenicol:
> Regarding the overall PR, do we want to:
> 
>     * Migrate grafana agent as well as the translation should be easy?
> 
>     * Migrate old promtail and grafana-agent configs to a template?
> 
>     * Try out default alloy pod discovery with structured metadata instead of the cumbersome config?

ooh, these are very valid questions.

So, currently we have:
- promtail for system logs, pod logs and audit logs
- grafana-agent for kube events

and we want to:
- replace grafana-agent with alloy
- replace promtail with alloy
- bonus: have less labels and use structured metadata instead

I'd love to go one step at a time, but given it requires olly-bundles releases, it would take quite some time so we have to plan in advance.
But it should be done in separate PRs anyway.

The most obvious plan to me would then be:
- This PR replaces promtail with grafana-agent, with a config for "alloy-logs".
- We could then add the kube-events config to it and remove the existing grafana-agent
- then look at structured metadata and how to migrate from relabellings to structured metadata.


Now, Quentin, you're saying there's a simpler/default config that would do it all with less config.
You're probably right, but the comparison is quite difficult when we change all at once.
Personnally, I'm more comfortable with this PR first, then a simplification (maybe breaking / changing some labels) later.

QuentinBisson:
Oh i'm not saying we should not have smaller PRs especially since this one is quite big already but we could have it done before the next olly bundle 1.4.0 release and v29 

I will need to review the promtail migrated config though :)

pkg/resource/logging-secret/reconciler.go
TheoBrigitte: I moved the secret creation and logic to select logging agent to `GenerateLoggingSecret`

pkg/resource/logging-agents-toggle/observability_bundle_configmap.go
TheoBrigitte: I added some logic within the `logging-agent-toggle` resource, to select the correct logging-agent. The info is then persisted within the `loggedcluster.Interface` object. This avoid to duplicate the logic to fetch and compare the observability-bundle version.

pkg/resource/logging-secret/alloy-logging-secret.go
TheoBrigitte: I killed this function

pkg/resource/logging-secret/alloy-logging-secret.go
TheoBrigitte: done

pkg/resource/logging-secret/alloy/logging-secret.yaml.template
TheoBrigitte: What do you mean ?

pkg/resource/logging-secret/alloy-logging-secret.go
TheoBrigitte: Moved this into the `init()` function

pkg/resource/logging-secret/alloy/logging-secret.yaml.template
QuentinBisson: I'm thinking most of what you put hère might bé more relevant as default in the alloy app than hère :)

pkg/resource/logging-secret/alloy-logging-secret.go
QuentinBisson: I was thinking about a New function but that works :)

pkg/resource/logging-secret/alloy/logging-secret.yaml.template
TheoBrigitte: I added those values as example into the Alloy app https://github.com/giantswarm/alloy-app/pull/21

TheoBrigitte:
Let's go one step at a time. Here we are "just" replacing the actual Promtail in place with Alloy and using a very similar config. We should first deploy and this and make sure everything works as expected (this was already tested but going live might always bring some surprises). Once we are ok with this. We can later thing about additional changes towards Kubernetes events and using metadata.

pkg/resource/logging-config/alloy-logging-config.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	loggedcluster "github.com/giantswarm/logging-operator/pkg/logged-cluster"
+)
+
+// GenerateAlloyLoggingConfig returns a configmap for
+// the logging extra-config
+// This is a plaholder function, in case we need to populate the configmap
+// for Alloy in the future.
+func GenerateAlloyLoggingConfig(lc loggedcluster.Interface) (string, error) {
```
QuentinBisson, line 11: Is there no way to only put secrets in the secret file as we did for promtail and have the config part here?

pkg/resource/logging-secret/alloy/logging-secret.yaml.template
QuentinBisson: Does this need to be a secret?

QuentinBisson: Especially since you mentino in the comments that this is a configmap :D

pkg/resource/logging-secret/alloy/logging-secret.yaml.template
```diff
@@ -0,0 +1,45 @@
+# This file was generated by logging-operator.
+# It configures Alloy to be used as a logging agent.
+# - configMap is generated from logging.alloy.template and passed as a string
+#   here and will be created by Alloy's chart.
+# - Alloy runs as a deamonset, with required tolerations in order to scrape logs
```
QuentinBisson, line 5:
```suggestion
# - Alloy runs as a daemonset, with required tolerations in order to scrape logs
```

pkg/resource/logging-config/alloy-logging-config.go
TheoBrigitte: There's no obvious way to achieve that, but I will attempt something in this direction

QuentinBisson (approved):
LGTM for now we can always improve :)

pkg/resource/logging-config/alloy-logging-config.go
TheoBrigitte: I moved the general config into the logging-config configmap, set the secret in logging-secret and pass them over to alloy using `envFrom`.

pkg/resource/logging-secret/alloy/logging-secret.yaml.template
TheoBrigitte: It's a configmap now

PR #167, opened 2024-07-19
Fix Alloy secret data and config templating

TheoBrigitte:

This PR fixes issues coming from https://github.com/giantswarm/logging-operator/pull/164

#### 1. Alloy app failed to be installed due to unexpected secret format;

```
parsing error: failed to parse `golem-logging-secret` in `org-giantswarm`, logs: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type map[string]interface {}
```

The logging-secret is used as extra config by the **app** (see [observability-bundle](https://github.com/giantswarm/observability-bundle/blob/c2fd32364b379a42aacefa76eaa841c8253f7777/helm/observability-bundle/values.yaml#L62)) and its being merged with logging-config by app operator to be used as values for Alloy app, since the secret is a key value pair app-operator fails to read it. But the expected behavior here is to have Alloy **pod** load a secret via [envFrom](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#secretenvsource-v1-core) in order for Alloy to be able to read those values from the environment and inject them into the configuration.

Therefore this PR:
- Set the logging-secret with an empty content
- Add alloy-secret resource to create a alloy-secret secret to be loaded by Alloy via envFrom.

#### 2. Alloy config templating went wrong due to a change from `text/template` package to `html/template`

This produced an invalid Alloy config with html encoded values:

```
      content: |-
        discovery.kubernetes &#34;kubernetes_pods&#34; {
                role = &#34;pod&#34;
        }
```

This PR fixes the import which fixes the templated config.


pkg/resource/logging-agents-toggle/observability_bundle_configmap.go
```diff
@@ -43,12 +44,13 @@ func GenerateObservabilityBundleConfigMap(ctx context.Context, lc loggedcluster.
...
+		appsToEnable[common.AlloyLogAgentName] = app{
 			Enabled: false,
 		}
-	case "alloy-logs":
-		appsToEnable["alloy-logs"] = app{
-			Enabled: true,
+	case common.AlloyLogAgentName:
+		appsToEnable[common.AlloyLogAgentName] = app{
+			Enabled:   true,
+			Namespace: common.AlloyLogAgentNamespace,
```
TheoBrigitte, line 53: Since this secret needs to live in the same namespace as Alloy pods, I am defining Alloy app namespace as a constant and using this for both creating the secret but also making sure Alloy is deployed in the given namespace in the logging-toggle resource.

pkg/common/common.go
```diff
@@ -21,6 +21,9 @@ const (
 	//#nosec G101
 	grafanaAgentExtraSecretName = "grafana-agent-secret"
 
+	AlloyLogAgentName      = "alloy-logs"
```
TheoBrigitte, line 24: This value is used for both the `--logging-agent` value and the Alloy app name in observability bundle.

QuentinBisson (approved):

TheoBrigitte:
Tested and working on `golem` :heavy_check_mark: 

image

see https://github.com/giantswarm/roadmap/issues/3524#issuecomment-2268960650

PR #168, opened 2024-07-20
Cache values in context

TheoBrigitte:
This PR propose a performance improvement by caching some values being used by multiple resources in the request context. It avoid those resources to be fetch from API mutliple times.
The caching is done the lazy way; within the function fetching those resources.

I would like someone with a broader overview of the operator and interaction between resources to confirm this does not break the current behaviour.
AFAIU the concerned resources are not modified within the request lifetime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Caches following values within the context to avoid fetching them multiple times:
+  - Observability bundle app version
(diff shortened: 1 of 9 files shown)


TheoBrigitte:
does not work, one cannot modify context in place

QuentinBisson:
Well you Can cache them in context if you update the context in the top of thé controller. But maybe you Can add them in thé logged cluster instead? Although i'm not sure cache is that useful because client-go should have a cache IIRC

TheoBrigitte:
True caching is already happening in client-go, but I wanted to reduce code duplication. But we can make this simpler and just move duplicated code in functions.

PR #170, opened 2024-07-21
Reduce boilerplate code realted to resource creation and update

TheoBrigitte:
This PR provides a solution to reduce the amount of duplicated code to create and update resources in the Kubernetes API.

It does so by using a generic function `common.EnsureCreatedOrUpdated` (suggestion welcome for naming).

This function encapsulates the logic to check if the resource exists and to create or update it. It can work with any object that satisfies the `client.Object` interface and takes the famous `needUpdate` function to decide whether the resource should be updated or not.

This was mainly motivated by a) reducing boilerplate and b) playing with generics in a real world scenario.

### DIFF
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -12,6 +12,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+// Generic constaint to handle pointer type.
+// This constaint allow to use a pointer to T as generic type.
+// This contraint inherits from client.Object as we expect
+// pointer to T to satisfy the client.Object interface.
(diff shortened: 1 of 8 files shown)


QuentinBisson (changes requested):
I would rather we use something like this https://github.com/kubernetes-sigs/controller-runtime/blob/12cc8d59fabe3a1dbbd927ae6aea077cf7962267/pkg/controller/controllerutil/controllerutil.go#L292 or Ask for advice in the controller-runtime slack Channel because I did not like this approach with PMO and I clearly don't liké it now 😅
I think the issue is we try to create too many différent resources if we start considering this is boilerplate

PR #183, opened 2024-08-09
fix alloy-secret naming

TheoBrigitte:
Fix alloy-secret naming

* Renaming secret to alloy-logging-secret
* Prefix secret with the cluster name
* Fix kyverno policy exception namespace

### DIFF
--- a/pkg/resource/alloy-secret/alloy-logging-secret.go
+++ b/pkg/resource/alloy-secret/alloy-logging-secret.go
@@ -51,7 +51,7 @@ func GenerateAlloyLoggingSecret(lc loggedcluster.Interface, credentialsSecret *v
 func SecretMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
 	metadata := metav1.ObjectMeta{
 		Name:      fmt.Sprintf("%s-%s", lc.GetClusterName(), secretName),
-		Namespace: lc.GetAppsNamespace(),
+		Namespace: common.AlloyLogAgentAppNamespace,
 		Labels:    map[string]string{},
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/resource/alloy-secret/alloy-logging-secret.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 const (
```
QuentinBisson, line 14: Is thé app not expecting clusterId-logging-secret? Why thé need to add alloy?

pkg/resource/logging-config/alloy/logging-config.yaml.template
```diff
@@ -46,3 +46,7 @@ alloy:
       - name: runlogjournal
         hostPath:
           path: /run/log/journal
+kyvernoPolicyExceptions:
+  enabled: true
+  # Ensure the kyverno policy exceptions are created in the namespace where alloy is installed
+  namespace: null
```
QuentinBisson, line 52: Why? I think they need to bé in the giantswarm Namespace for now
```suggestion
  namespace: null
```

pkg/resource/alloy-secret/alloy-logging-secret.go
TheoBrigitte: Yes the final format is `$clusterID-alloy-logging-secret`. I added the `logging` part to make it explicit that this secret is intended for logging usage.

pkg/resource/logging-config/alloy/logging-config.yaml.template
TheoBrigitte: You're right this is not needed, removing it.

QuentinBisson (approved):

PR #194, opened 2024-08-20
fix-alloy-logs-security-context

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/31179

Replicating https://github.com/giantswarm/alloy-app/pull/36/files in logging operator

@TheoBrigitte see why I did not like it having it all here? :D


pkg/resource/logging-config/alloy/logging-config.yaml.template
```diff
@@ -28,10 +28,20 @@ alloy:
...
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop:
+        - ALL
+      readOnlyRootFilesystem: true
+      runAsUser: 473
+      runAsGroup: 473
+      runAsNonRoot: true
+      seccompProfile:
+        type: RuntimeDefault
```
TheoBrigitte, line 44: This won't work.

Alloy needs to access filesystem as root in order to be able to read log files.
I am unsure about the `readOnlyRootFilesystem` flag, this would need to be tested.

Also adding the `/tmp/alloy` might have impact on the memory, please test this change.

TheoBrigitte (changes requested):

pkg/resource/logging-config/alloy/logging-config.yaml.template
QuentinBisson: I am testing this now but this should work fine because alloy docker image is built-in with a non root user (473) https://grafana.com/docs/alloy/latest/configure/nonroot/.

/tmp/alloy is the default config, we need an empty dir because it's the only required fine il the root path that breaks the readOnlyRootFilesystem. I tested the grafana agent PR with that and it works so I'm trying it with alloy but it's roughtly similar

QuentinBisson: Now it only configures read only filesystem which has no impacts on resource usage and the number of logs on loki

QuentinBisson:
@TheoBrigitte I kept root user but removed the capabilities we do not need

QuantumEnigmaa (approved):
LGTM if tested and working

TheoBrigitte (approved):

PR #201, opened 2024-08-26
add chart testing to ci pipeline

QuantumEnigmaa:
Towards https://github.com/giantswarm/roadmap/issues/3616

e2e tests will be added in a a separate PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,8 +9,10 @@ jobs:
       - image: giantswarm/helm-chart-testing:v3.11.0
     steps:
       - checkout
-      - run: cd helm/logging-operator && helm dep up
-      - run: helm template helm/logging-operator
+      - run: |-
+          cd helm/logging-operator
+          helm dep up
+          helm template helm/logging-operator
 
 
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
       - run: |-
           cd helm/logging-operator
           helm dep up
-          helm template helm/logging-operator
+          helm template .
 
 
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,16 @@ version: 2.1
 orbs:
   architect: giantswarm/architect@5.5.2
 
+jobs:
+  template-chart:
+    docker:
+      - image: giantswarm/helm-chart-testing:v3.11.0
+    steps:
+      - checkout
+      - run: cd helm/logging-operator && helm dep up
+      - run: helm template helm/logging-operator
+
+
 workflows:
   build:
     jobs:
@@ -12,6 +22,15 @@ workflows:
           filters:
             tags:
               only: /^v.*/
+      
+      - template-chart:
+          name: "template-chart"
+          filters:
+            branches:
+              ignore:
+                - main
+          requires:
+            - "go-build"
 
       - architect/push-to-registries:
           context: architect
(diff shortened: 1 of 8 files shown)


QuentinBisson (approved):

.circleci/config.yml
```diff
@@ -3,6 +3,16 @@ version: 2.1
...
   architect: giantswarm/architect@5.6.0
 
+jobs:
+  template-chart:
+    docker:
+      - image: giantswarm/helm-chart-testing:v3.11.0
+    steps:
+      - checkout
+      - run: cd helm/logging-operator && helm dep up
+      - run: helm template helm/logging-operator
```
TheoBrigitte, line 13: This could be a single run I believe
```suggestion
      - run: |-
          cd helm/logging-operator
          helm dep up
          helm template helm/logging-operator
```

tests/test-values.yaml
```diff
@@ -0,0 +1,8 @@
+lobal:
```
TheoBrigitte, line 1:
```suggestion
global:
```

tests/ats/test_logging_op.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    Testing apiserver availability.
+    """
+    assert kube_cluster.kube_client is not None
+    assert len(pykube.Node.objects(kube_cluster.kube_client)) >= 1
+
+# scope "module" means this is run only once, for the first test case requesting! It might be tricky
+# if you want to assert this multiple times
+@pytest.fixture(scope="module")
+def ic_deployment(kube_cluster: Cluster) -> List[pykube.Deployment]:
```
TheoBrigitte, line 27: Could you please add a description on what we are testing here.

tests/ats/test_logging_op.py
```diff
@@ -0,0 +1,63 @@
...
+                       'app.kubernetes.io/name': 'logging-operator', 'app.kubernetes.io/instance': 'logging-operator'})
+
+    return pods
+
+# when we start the tests on circleci, we have to wait for pods to be available, hence
+# this additional delay and retries
+@pytest.mark.smoke
+@pytest.mark.upgrade
+@pytest.mark.flaky(reruns=5, reruns_delay=10)
+def test_pods_available(ic_deployment: List[pykube.Deployment]):
```
TheoBrigitte, line 60: Could you please add a description on what we are testing here.

QuantumEnigmaa: I added some comments explaining what's being tested for each :)

tests/ats/test_logging_op.py
QuantumEnigmaa: I added some comments explaining what's being tested for each :)

QuantumEnigmaa:
@TheoBrigitte All good for you ?

hervenicol:
IIUC this does 3 runs of tests:
- smoke tests
- functional tests
- upgrade tests

It seems to me that the upgrade tests are a subset of the smoke tests, and I didn't see the functional tests.

Maybe it's because I'm not used to reading these tests, but is there a way to make it easier to understand?

QuantumEnigmaa:
> I didn't see the functional tests.

That's normal : if you look at the `.ats/main.yaml` file which is configuring ATS, there's this piece of config : 
```
skip-steps: [functional]
```
Originally this is was just a leftover from the source repo I first copied the tests from but eventually I though that in in any case, we would most likely don't need any functionnal tests as to ensure that the operator's functions are working as expected, we need it to be deployed in a "prod-like" environment. So I'm leaving those kind of checks to the e2e tests. 

> is there a way to make it easier to understand?

I'm not sure how but I could maybe add some additional comments or a README in the `ats` directory explaining which tests are performed and why.

hervenicol:
> > is there a way to make it easier to understand?
> 
> I'm not sure how but I could maybe add some additional comments or a README in the `ats` directory explaining which tests are performed and why.

I was thinking maybe splitting tests into multiple files?
With one file with all the common functions, one file with smoke tests and one with upgrade tests.
Even though they'd call the same functions, it may make it easier to follow a scenario.

Not sure it's a good idea though, it should probably be tested in a separate PR as a refactoring of this one. Once this one is merged.

QuantumEnigmaa:
Hmmm personnaly I think I would be even more lost when it would come to understanding the tests. I would agree if we had way more tests performed but in the present case with only 3 actual tests, I think it's still better to have those in a single file.

hervenicol (approved):
LGTM.

PR #215, opened 2024-09-05
Write all to giantswarm tenant

QuentinBisson:
Towards https://github.com/giantswarm/roadmap/issues/3726

This PR is the start of the move to multi-tenancy for logs. For the new datasources to work, we need to send all logs to the giant swarm tenant. 

This might impact Loki on some installations which we might need to scale up (either the number of buckets or the pod resources)

Once this PR is merged, we can try to work on supporting self-service tenancy on pod logs?


QuentinBisson:
Trying this out on grizzly

TheoBrigitte (commented):
What is the impact of this ? And why do we need it ?

QuentinBisson:
Oh I did not write anything ...

So once we have the multi-tenancy POC running, we will most likely have giantswarm logs under 1 tenant only so the impact is that loki will most likely not like it (need to be scaled) on some installations with lots of traffic but this would allow us to create alerts based on logs on MCs and WCs alike when now it only works on MCs

QuentinBisson:
@giantswarm/team-atlas I edited the top description. I forced pushed because the original main branch was too far behind

pkg/common/common.go
```diff
@@ -14,6 +14,8 @@ import (
 const (
 	// ReadUser is the global user for reading logs
 	ReadUser = "read"
+	// DefaultWriteTenant is the default tenant for writing logs
+	DefaultWriteTenant = "giantswarm"
```
QuantumEnigmaa, line 18: Aren't we renaming the default tenant to "Shared Org" with olly-op ? Or am I confusing things here ? :sweat_smile:

QuentinBisson: Main Org will be renamed yes but not the tenant :D

QuantumEnigmaa: Ok my bad then :)

QuantumEnigmaa (approved):
LGTM

QuantumEnigmaa (approved):

TheoBrigitte (approved):

QuentinBisson:
we're officially out of freeze

PR #221, opened 2024-09-17
Fix Alloy logs secret handling

TheoBrigitte:

Fixes an issue where the current secret is always created on the MC which makes it impossible to configure Alloy logs on a WC.

  - Pass secret environment variables via helm chart values
    using the extraSecretEnv directive
  - Ensure old secret is deleted

### DIFF
--- a/pkg/resource/alloy-secret/reconciler.go
+++ b/pkg/resource/alloy-secret/reconciler.go
@@ -2,20 +2,15 @@ package alloysecret
 
 import (
 	"context"
-	"fmt"
 	"reflect"
-	"time"
 
-	appv1 "github.com/giantswarm/apiextensions-application/api/v1alpha1"
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	apimachineryerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
-	"github.com/giantswarm/logging-operator/pkg/common"
 	loggedcluster "github.com/giantswarm/logging-operator/pkg/logged-cluster"
-	loggingcredentials "github.com/giantswarm/logging-operator/pkg/resource/logging-credentials"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


pkg/resource/alloy-secret/reconciler.go
```diff
@@ -29,81 +23,12 @@ type Reconciler struct {
...
-
-	// Retrieve Loki ingress name
-	lokiURL, err := common.ReadProxyIngressURL(ctx, lc, r.Client)
-	if err != nil {
-		return ctrl.Result{}, errors.WithStack(err)
-	}
-
-	// Get desired secret
-	desiredAlloySecret, err := GenerateAlloyLoggingSecret(lc, &loggingCredentialsSecret, lokiURL)
+	result, err := r.ReconcileDelete(ctx, lc)
```
QuentinBisson, line 26: Do you want to delete always?

QuentinBisson: Oh yes you do

pkg/resource/logging-secret/alloy-logging-secret.go
```diff
@@ -1,13 +1,69 @@
...
+	//go:embed alloy/logging-secret.yaml.template
+	alloySecret         string
+	alloySecretTemplate *template.Template
+)
+
+func init() {
+	alloySecretTemplate = template.Must(template.New("logging-secret.yaml").Funcs(sprig.FuncMap()).Parse(alloySecret))
+}
+
 func GenerateAlloyLoggingSecret(lc loggedcluster.Interface, credentialsSecret *v1.Secret, lokiURL string) (map[string][]byte, error) {
```
QuentinBisson, line 35: Are you not missing a call to this function in the controller?

QuentinBisson: Nevermind

pkg/resource/alloy-secret/reconciler.go
TheoBrigitte: Yes for now this should always be deleted. In the next release we can remove this resource entirely, I'll add a TODO about this.

pkg/resource/logging-secret/alloy-logging-secret.go
TheoBrigitte: This function is already being called here https://github.com/giantswarm/logging-operator/blob/070de870fd41dcaff7e1903406c4b115e32c937b/pkg/resource/logging-secret/logging-secret.go#L30

QuentinBisson:
Alright so you fixed 2 issues by refactoring a bit, the first is that the initial alloy secret was created in the kube-system namespace so it would not work for WC apps and the second is that you ensured the secret is now mounted in the alloy app via the extraSecretEnv stuff right?

It's not your fault but moving code from a reconciler to another is not easy to read and that's why I like my business logic separated from the infrastructure code :)

QuentinBisson (approved):
Overall looks good to me

pkg/resource/logging-secret/alloy-logging-secret.go
QuentinBisson: Yes sorry I had to dig deeper :)

PR #233, opened 2024-09-26
Configure Alloy to be use as a self-service tool for logs (source podlogs)

TheoBrigitte:

This PR add the capability to dynamically configure log targets using the `giantswarm.io/logging` label. This label can be used on Pods and Namespaces. This label can take 2 distinct values `true` or `false`, when the label is not set its considered `false`.
- Pods in any namespace can be labeled
- Namespace can be labeled, the value used for the namespace will be the default for all pods in this namespace
- Pods label take precedence over Namespace label, e.g. logging can be enabled for a pod in a namespace where logging is disabled and vice versa.

<details>
<summary>Examples</summary>

||Pod `logging=true`|Pod `logging=false`|Pod `logging unset`
|-|------------------------------|--------------------------------|--------------------------
Namespace `logging=true`|:heavy_check_mark:|:heavy_multiplication_x:|:heavy_check_mark:
Namespace `logging=false`|:heavy_check_mark:|:heavy_multiplication_x:|:heavy_multiplication_x:
Namespace `logging unset`|:heavy_check_mark:|:heavy_multiplication_x:|:heavy_multiplication_x:

- `logging=true` : short version for `giantswarm.io/logging=true` label
- `logging=false`: short version for `giantswarm.io/logging=false` label
- `logging unset`: short version for no `giantswarm.io/logging` label set

</details>

This is achieved by using `PodLogs` to discover targets, which lead to log lines being tailed through Kubernetes API server instead of fetching them from disk, this has some performance impact being describe here https://github.com/giantswarm/roadmap/issues/3518#issuecomment-2391961917. [`clustering`](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.source.podlogs/#clustering-block) is enabled in order to spread the load across the different Alloy pods/nodes.

The new PodLogs resources added do respect the previous default behaviour:
- Fetch logs for all pods on all namespaces on MC
- Fetch logs for all pods in kube-system and giantswarm namespaces on WC

### Alloy graph before
image

### Alloy graph after
image


pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
```diff
@@ -58,3 +61,60 @@ alloy:
           path: /run/log/journal
       - name: alloy-tmp
         emptyDir: {}
+podLogs:
+{{- if .IsWorkloadCluster }}
+- name: namespace-default
+  namespace: giantswarm
```
QuentinBisson, line 67:
```suggestion
  namespace: giantswarm
```

Maybe this should bé in thé kube-system.namespace? Giantswarm namespace will soon disappear on WCs

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
```diff
@@ -58,3 +61,60 @@ alloy:
...
+  spec:
+    selector: {}
+    namespaceSelector:
+      matchExpressions:
+      - key: kubernetes.io/metadata.name
+        operator: In
+        values:
+        - kube-system
```
QuentinBisson, line 75: @giantswarm/team-shield i'm guessing WE should add kyverno namespace so.we Can collect the kyverno logs?

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
```diff
@@ -58,3 +61,60 @@ alloy:
...
+  spec:
+    selector:
+      matchLabels:
+        giantswarm.io/logging: enabled
+    namespaceSelector:
+      matchExpressions:
+      - key: giantswarm.io/logging
+        operator: NotIn
+        values:
+        - enabled
```
QuentinBisson, line 88: Could WE not simple use giantswarm.io/logging: true liké we do for clusters?

QuentinBisson: Did you test this with logging disabled at the cluster level ?

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
```diff
@@ -58,3 +61,60 @@ alloy:
...
+        operator: In
+        values:
+        - enabled
+      - key: kubernetes.io/metadata.name
+        operator: NotIn
+        values:
+        - kube-system
+        - giantswarm
+{{- else }}
+- name: pod-selector
```
QuentinBisson, line 115: We probably want to use the same stuff on MCs for dogfooding so we need to create an issue for this

pkg/resource/logging-config/alloy/logging.alloy.template
```diff
@@ -1,9 +1,20 @@
-discovery.kubernetes "kubernetes_pods" {
-	role = "pod"
+// Kubernetes pods logs
```
QuentinBisson, line 1: There's no notion of tenancy right?

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
TheoBrigitte: Ok that makes sense, I'll move PodLogs into kube-system namespace

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
TheoBrigitte: Good point, I haven't tested it with logging disabled at cluster level.

Sure we can use true/false instead of enabled/disabled

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
TheoBrigitte: I would like to keep this for later, as this include already a lot of changes I would like to stay as close as possible to the current setup.

pkg/resource/logging-config/alloy/logging.alloy.template
TheoBrigitte: Nope

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
stone-z: > Giantswarm namespace will soon disappear on WCs

Really? I didn't know this :open_mouth: are there some docs about it?

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
stone-z: Yes it would be super useful to have the Kyverno logs :slightly_smiling_face: 
What did Atlas have in mind about ownership/config for that label? I would say it might make sense to have this list of namespaces set through `shared-configs` or something similar instead of being baked into logging-operator, but we could also set it inside the Kyverno app if Atlas wanted to leave it up to the teams

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
QuentinBisson: Yes that's why I said to not forget creating a new issue 😂

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
QuentinBisson: The idea is that we would set a static list for well static namespace liké kyverno and kube-system and the rest would bé for customers to set up using labels. Maybe this could bé externalités via config sure

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
QuentinBisson: Nope but it's coming with the removal of the chart-operator on WCs

QuentinBisson: The goal is to not have any giant swarm code on WCs

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
TheoBrigitte: Yes I agree we should move this list out of logging-operator

helm/logging-operator/values.yaml
```diff
@@ -38,6 +38,7 @@ networkPolicy:
   flavor: cilium
 
 loggingOperator:
+  defaultWorkloadClusterNamespaces: "kube-system,giantswarm"
```
QuentinBisson, line 41: Sure the name is fine but we will probably change it to defaultNamespaces or defaultStaticNamespaces later right?

main.go
```diff
@@ -65,7 +66,19 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+type StringSliceVar []string
```
QuentinBisson, line 69: Why is that not built in??

QuentinBisson:
Do we need to run it as a daemonset if we use pod logs? I think we should be able to use a deployment? And I guess we need less permissions (no root access anymore) right?

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
TheoBrigitte: Issue created: https://github.com/giantswarm/roadmap/issues/3715

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
TheoBrigitte: PodLogs are now created in kube-system namespace

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
TheoBrigitte: I added the following Helm chart value, this can be overwritten via usual mechanism like shared-configs
```
loggingOperator:
  defaultWorkloadClusterNamespaces: "kube-system,giantswarm"
```

helm/logging-operator/values.yaml
TheoBrigitte: I though `defaultNamespaces` was confusing because this will only affect workload clusters.

QuentinBisson: For now yes but not once https://github.com/giantswarm/roadmap/issues/3715 is done :)

main.go
TheoBrigitte: There is something like this in the [`spf13/pflag`](https://pkg.go.dev/github.com/spf13/pflag#StringSliceVar) library, but this does not work, but this does not works with controller-runtime `BindFlags` function :shrug:

TheoBrigitte:
> Do we need to run it as a daemonset if we use pod logs? I think we should be able to use a deployment? And I guess we need less permissions (no root access anymore) right?

We could run Alloy as deployment, but yet we would still need to run an Alloy as daemonset to grab systemd and audits logs, which would result in splitting this into 2 Alloy which is a lot more work since logging-operator would need to configure 2 different Alloy. I would rather that we keep everything in the same Alloy.

QuentinBisson:
We might need to change the alloy version that triggers the change right?

TheoBrigitte:
> We might need to change the alloy version that triggers the change right?

What do you mean ?

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
TheoBrigitte: I changed the values to true/false.

Also tested with logging disabled at the cluster level, Alloy is uninstalled so it works :heavy_check_mark:

TheoBrigitte:
> I'm thinking about this
> 
> https://github.com/giantswarm/logging-operator/blob/dac44d06e6916370d4559184920d6acc77f750cf/pkg/resource/logging-agents-toggle/reconciler.go#L36
> but maybe we do not care

Yes I still need to bump this once Alloy and observability-bundle new versions have been released

QuentinBisson (approved):

helm/logging-operator/values.yaml
TheoBrigitte: Renamed it to `defaultNamespaces`

TheoBrigitte:
A different Alloy config is now generated when observability-bundle is below or above/equal to 1.7.0.

main.go
```diff
@@ -75,6 +88,7 @@ func main() {
 	var profilesAddr string
 	var probeAddr string
 	var vintageMode bool
+	flag.Var(&defaultNamespaces, "default-namespaces", "List of namespaces to scrapes logs from by default on workload clusters")
```
QuentinBisson, line 91:
```suggestion
	flag.Var(&defaultNamespaces, "default-namespaces", "List of namespaces to collect logs from by default on workload clusters")
```

pkg/resource/logging-config/alloy-logging-config_test.go
```diff
@@ -0,0 +1,114 @@
+package loggingconfig
+
+import (
```
QuentinBisson, line 3: I really hate those golden files :(

QuentinBisson: I would prefer having 2 templates.based on the version

QuentinBisson: Anyway, if I'm the only one that complaints then at least remove the update function. I want people to have to fix their template otherwise the golden files are used for PR reviews and not testing and they should not bé hère but in the PR description

TheoBrigitte: This part of the code becomes quite complex and we need a way to ensure we are not introducing unwanted changes. The update function makes it much easier to update tests, it should then be on the reviewer to asses that the changes are correct, we should not make our life harder by requiring to change those files manually.

pkg/resource/logging-config/alloy-logging-config_test.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+		{
+			goldenFile:                 "alloy/test/logging-config.alloy.170_MC.yaml",
+			observabilityBundleVersion: "1.7.0",
+			defaultNamespaces:          []string{"test-selector"},
+			installationName:           "test-installation",
+			clusterName:                "test-installation",
+		},
+		{
+			goldenFile:                 "alloy/test/logging-config.alloy.170_WC.yaml",
```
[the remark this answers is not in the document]
TheoBrigitte: Config for WC and MC are different for now, we will changes this once we have the same config for both.

hervenicol:
I don't see anything regarding multi-tenancy. Does it mean all logs are sent to the default orgid?

TheoBrigitte:
> I don't see anything regarding multi-tenancy. Does it mean all logs are sent to the default orgid?

Yes I removed the multi-tenancy configuration for now as I could not make it work. Should we add that ? I would need some help on this then.

hervenicol:
Well, I'm not sure multi-tenancy is needed for a first iteration, but I'm quite confident it's needed before our customers would use this feature.
So maybe this should be an extra step in https://github.com/giantswarm/roadmap/issues/3518 ?

QuentinBisson:
Maybe they should be sent to the cluster_id tenant for now and we revisit once the multi-tenancy epic is closed. This should be revisited once we start with multi-tenancy for logs right @Rotfuks ?

TheoBrigitte:
Going with this and we can treat the multi-tenancy and tenant override later on then.

PR #235, opened 2024-10-04
Configure Alloy to be used as a self-service too for logs (source files)

TheoBrigitte:


pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
```diff
@@ -44,6 +47,22 @@ alloy:
         type: RuntimeDefault
   controller:
     type: daemonset
+    extraContainers:
+      - name: alloy-namespaces
+        image: bitnami/kubectl:latest
```
TheoBrigitte, line 52: Change this to use a giantswarm image

TheoBrigitte:
Still needs to implement the tenant name override from label, like its done here https://github.com/giantswarm/logging-operator/pull/233/files#diff-405f451506c2146b4cf915863a848d9a42f39a3b292a9b6f9ada78b4eac32598R85

TheoBrigitte:
Not following this route as it would not allow to override the tenant when using labels on namespaces, due to the namespaces labels not being available when using discovery.kubernetes.

Trying to discover those labels using a sidecar container also does not work, the idea was to fetch the namespace label and produce a list of `namespace_A:tenant_name_A;namespace_B:tenant_name_B;` and then extract the tenant name using the following relabel rule
```
rule {
        source_labels = concat(["__meta_kubernetes_namespace"], split(local.file.namespaces_tenant.content, ";"))
       regex         = "(.+);.*(\1):([0-9]+);.*"

        target_label  = "tenant_namespace"
}
```

Unfortunately there is no support for backreference (`\1`) in Alloy/Prometheus regex [RE2 engine](https://github.com/google/re2/wiki/Syntax)

PR #241, opened 2024-10-15
Use PodLogs only for log targets discovery instead of labels

TheoBrigitte:

This PR updates the current implementation to select log targets only using PodLogs resources and drop the predefined PodLogs used for label selection.

Other changes:

- Select all PodLogs
- Renamed our PodLogs to `all-pods` on MC and `default-namespaces` on WC to be more explicit
- Fixed empty podLogs directive when no default namespaces are provided


pkg/resource/logging-config/alloy/logging.alloy.template
```diff
@@ -4,8 +4,9 @@ loki.source.podlogs "kubernetes_pods" {
   forward_to      = [loki.relabel.kubernetes_pods.receiver]
 
   selector {
-    match_labels = {
-      "giantswarm.io/managed-by" = "alloy-logs",
+    match_expression = {
+      key      = "giantswarm.io/managed-by"
```
QuentinBisson, line 8: Why do we need this label? This means we cannot discover other pod lods right?

TheoBrigitte: Right, maybe we should keep it to `giantswarm.io/managed-by: alloy-logs` so it's clear those PodLogs are going through our alloy-logs app.

QuentinBisson: But that means customers need to add giantswarm.io/managed-by so their pod logs are found right? Why not get all pod logs?

QuentinBisson: Also, if we were to use a label we should use those https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels and not giantswarm custom ones

QuentinBisson: Honestly, I think we're building limitations we don't need, or at least not yet. We discussed this morning that we did not want to go label based and now we need to add a label on the pod log anyway. I'm not sure I see value considering we most likely have no customer running alloy. @giantswarm/team-atlas what do you think? 

Also, if we were to use a label, we might as well use the same one we use for service monitors right?

TheoBrigitte: I removed this selector

QuentinBisson (approved):
Don't forget to update the PR description :)

QuantumEnigmaa (approved):
LGTM

PR #251, opened 2024-10-30
disable grafana-agent pod logs crds as they are causing issues with alloy-logs

QuentinBisson:

This PR disables the grafana-agent pod logs as they are conlflicting with the alloy-logs crds and helm does not like that at all:

```"level":"info","ts":"2024-10-29T10:50:41Z","msg":"App status for 't-vtzyp1tjrlyibifczu-grafana-agent' is not yet as expected: expectedStatus='deployed' actualStatus='already-exists' (reason: 'object already exists: (rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition \"podlogs.monitoring.grafana.com\" in namespace \"\" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key \"app.kubernetes.io/managed-by\": must be set to \"Helm\"; annotation validation error: missing key \"meta.helm.sh/release-name\": must be set to \"grafana-agent\"; annotation validation error: missing key \"meta.helm.sh/release-namespace\": must be set to \"kube-system\")')"}```

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  architect: giantswarm/architect@5.11.0
+  architect: giantswarm/architect@5.11.1
 
 jobs:
   template-chart:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Reconcile clusters when the observability bundle version changes.
+
 ### Fixed
 
 - Disable crd installation from grafana agent as this is causing issues with the new v29 releases.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Disable crd installation from grafana agent as this is causing issues with the new v29 releases.
+
 ## [0.14.0] - 2024-10-29
 
 ### Changed
--- a/internal/controller/predicates/predicates.go
+++ b/internal/controller/predicates/predicates.go
@@ -36,14 +36,13 @@ func (ObservabilityBundleAppVersionChangedPredicate) Update(e event.UpdateEvent)
 		return false
 	}
 
-	oldAppVersion, err := semver.New(oldApp.Spec.Version)
+	oldAppVersion, err := semver.Parse(oldApp.Spec.Version)
 	if err != nil {
 		return false
 	}
-	newAppVersion, err := semver.New(newApp.Spec.Version)
+	newAppVersion, err := semver.Parse(newApp.Spec.Version)
 	if err != nil {
 		return false
 	}
-	breakingVersion := semver.MustParse("1.0.0")
-	return oldAppVersion.LT(breakingVersion) && newAppVersion.GTE(breakingVersion)
+	return oldAppVersion.NE(newAppVersion)
 }
--- a/pkg/resource/grafana-agent-config/grafana-agent-config.go
+++ b/pkg/resource/grafana-agent-config/grafana-agent-config.go
@@ -30,6 +30,7 @@ type values struct {
 type grafanaAgent struct {
 	Agent      agent      `yaml:"agent" json:"agent"`
 	Controller controller `yaml:"controller" json:"controller"`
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

**Comment by hervenicol** [comment]
According to the CI it breaks logging-operator deployment? Maybe it does not like that it has no podlogs CRDs?

So, the plan is to disable podlogs CRDs installation from both grafana-agent and alloy-metrics, and let alloy-logs install them, right?
Is the podlogs CRD the only CRD deployed by alloy? Are we sure there won't be another CRD required by grafana-agent or alloy-metrics?

What happens to existing clusters when we disable CRDs from one app, won't it uninstall them?
Anyways CRDs ownership will need to change, and we have to manually do it, right? Meaning that's a manual action to run when clusters will get upgraded?

Also, we want grafana-agent to skip CRDs installation only on clusters running alloy-logs I guess, meaning it should still deploy CRDs on older clusters? Or maybe we don't use them on older clusters so we don't care?

**Comment by QuentinBisson** [comment]
> According to the CI it breaks logging-operator deployment? Maybe it does not like that it has no podlogs CRDs?

So the CI breaks because of an incorrect pytest-helm-chart version that hangs forever, this is why all our apps with ats fail are failing their new architect pr

> So, the plan is to disable podlogs CRDs installation from both grafana-agent and alloy-metrics, and let alloy-logs install them, right?
Yes

> Is the podlogs CRD the only CRD deployed by alloy? Are we sure there won't be another CRD required by grafana-agent or alloy-metrics?

Yes it's the only one at the moment
 
> What happens to existing clusters when we disable CRDs from one app, won't it uninstall them?

So yes it should be first we don't really have any clusters using pod logs now as they are expected to be in 29.4 or v30 and we have none yet in the latest version. If it is the case, it's only test clusters anyway

> Anyways CRDs ownership will need to change, and we have to manually do it, right? Meaning that's a manual action to run when clusters will get upgraded?

With the fix, all existing PodLog CRD managed by the grafana-agent will be removed so we will not have to do anything

> Also, we want grafana-agent to skip CRDs installation only on clusters running alloy-logs I guess, meaning it should still deploy CRDs on older clusters? Or maybe we don't use them on older clusters so we don't care?

We do not use them on old clusters, we only use them with alloy-logs so we don't care at all :D

**Comment by QuentinBisson** [comment]
Pytest-helm-chart was an old version

**Comment by QuentinBisson** [comment]
Tests are fixed @hervenicol :)

**Comment by QuentinBisson** [comment]
Created a v28 capa cluster.

Charts deployed:

```
k get charts -n giantswarm
NAME                           VERSION      LAST DEPLOYED   STATUS
aws-ebs-csi-driver-smons       0.1.0        9m10s           deployed
aws-pod-identity-webhook       1.16.0       9m25s           deployed
capi-node-labeler              0.5.0        14m             deployed
cert-exporter                  2.9.0        8m22s           deployed
cert-manager                   3.7.9        9m51s           deployed
chart-operator                 3.3.0        14m             deployed
chart-operator-extensions      1.1.2        8m31s           deployed
cilium-servicemonitors         0.1.2        9m27s           deployed
cluster-autoscaler             1.28.5-gs1   8m32s           deployed
etcd-k8s-res-count-exporter    1.10.0       9m27s           deployed
external-dns                   3.1.0        9m35s           deployed
grafana-agent                  0.4.3        9m8s            deployed
irsa-servicemonitors           0.0.1        9m9s            deployed
k8s-audit-metrics              0.9.0        9m25s           deployed
k8s-dns-node-cache             2.6.2        9m6s            deployed
kube-prometheus-stack          9.1.2        9m24s           deployed
kyverno                        0.17.14      10m             deployed
kyverno-policies               0.20.2       9m11s           deployed
kyverno-policy-operator        0.0.7        8m31s           deployed
metrics-server                 2.4.2        9m9s            deployed
net-exporter                   1.19.0       9m11s           deployed
node-exporter                  1.19.0       9m11s           deployed
prometheus-agent               0.6.9        9m7s            deployed
prometheus-blackbox-exporter   0.4.1        9m26s           deployed
prometheus-operator-crd        9.0.0        11m             deployed
promtail                       1.5.3        9m12s           deployed
teleport-kube-agent            0.9.0        14m             deployed
vertical-pod-autoscaler        5.2.2        9m6s            deployed
```

I added a podlog CR:

```
k get podlogs -A
NAMESPACE     NAME              AGE
kube-system   example-podlogs   82s
```

Upgrading to v29.3.0

hervenicol (approved):
We checked that the annotated grafana-agent-deployed CRDs are deleted with this config, then any alloy may create their CRDs.
The alloy-deployed CRDs won't create the same conflict as they don't have ownership metadata.

So, LGTM :+1:

PR #254, opened 2024-10-31
force reconciliation whenever the observability-bundle version changes

QuentinBisson:

This PR fixes the current cluster upgrade issues that has been happening in releases since we enabled alloy as our default logging agent.

For context, here is how a cluster upgrade goes:

1. Upgrade Cluster CR
2. All operators act on it including the logging-operator which generates the config for the currently deployed observability version

The issue here is that in when we upgrade capv releases from v28 to v29, we upgrade the observability bundle from a version < 1.5.3 to a version >= 1.5.3 which is when we deploy alloy as our logging agent: https://github.com/giantswarm/logging-operator/blob/1e338f926451a65f456d00cc155366631ffc13d5/pkg/resource/logging-agents-toggle/observability_bundle_configmap.go#L36

As controllers run in parallel, this causes a race condition between the component that deploys the new app versions and the logging-operator which acts faster so the generated config targets an old observability-bundle version that does not use alloy and so does not create the secret alloy needs to start https://github.com/giantswarm/logging-operator/blob/1e338f926451a65f456d00cc155366631ffc13d5/pkg/resource/logging-secret/logging-secret.go#L30

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Reconcile clusters for any observability bundle version change after v1.0.0 instead of only when we upgrade to v1.0.0 because app upgrades happen way after cluster changes when upgrading a cluster.
+
 ## [0.14.0] - 2024-10-29
 
 ### Changed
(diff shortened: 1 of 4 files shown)


QuantumEnigmaa (approved):
LGTM

hervenicol (approved):
LGTM

PR #257, opened 2024-11-02
fix failing release upgrades

QuentinBisson:

Fixes issues when running observability bundle 1.5.3 as it does not support the [new secret configuration mechanism ](https://github.com/giantswarm/alloy-app/blob/main/CHANGELOG.md#040---2024-08-12)added in observability bundle [1.6.0](https://github.com/giantswarm/observability-bundle/blob/main/CHANGELOG.md#160---2024-08-20). This causes clusters with version 1.5.3 and alloy enabled to fail

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


hervenicol (approved):
Too bad we can't support earlier olly-bundles. We probably have to update customer communication because this will delay self-service logging.

PR #261, opened 2024-11-07
add tenancy in self-service

QuentinBisson:

This PR allows customers to set the tenant id they want to use for their logs in two different ways:
- They can add a podlog that matches any pods that they want and set the tenant via a relabeling like so:
```yaml
- apiVersion: monitoring.grafana.com/v1alpha2
  kind: PodLogs
  metadata:
    name: test-monitoring
    namespace: monitoring
  spec:
    namespaceSelector:
      matchExpressions:
      - key: kubernetes.io/metadata.name
        operator: In
        values:
        - monitoring
    relabelings: # This is how to set the tenant that will be processed by the pipeline
    - action: replace
      replacement: mytenant
      targetLabel: giantswarm_observability_tenant
    selector: {}
```
- They can also add the `observability.giantswarm.io/tenant` label with the tenant as a key in the pod labels and those logs will automatically be scraped by our existing pipeline (c.f. the new customer-logs pod log in this PR)

### Checklist


pkg/resource/logging-config/alloy/logging.alloy.template
```diff
@@ -292,7 +302,7 @@ loki.write "default" {
 	endpoint {
 		url                = env("{{ .LokiURLEnvVarName }}")
 		max_backoff_period = "{{ .MaxBackoffPeriod }}"
-		tenant_id          = env("{{ .TenantIDEnvVarName }}")
+		tenant_id          = "giantswarm"
```
TheoBrigitte, line 305: I liked the idea of using an env var to set the default tenant, why removing this ?

QuentinBisson: What value does it bring?

pkg/resource/logging-config/alloy/logging.alloy.template
```diff
@@ -84,7 +84,7 @@ discovery.relabel "kubernetes_pods" {
 		target_label  = "container"
 	}
 
-{{- if not .SupportPodLogs }}
+  {{- if not .SupportPodLogs }}
```
QuentinBisson, line 87: Needs to be update to match with https://github.com/grafana/alloy/issues/276

pkg/resource/logging-config/alloy/logging.alloy.template
```diff
@@ -89,7 +89,25 @@ discovery.relabel "kubernetes_pods" {
...
+		target_label  = "__tenant_id__"
+	}
+
+	// We drop the label to not have it ingested in loki
+	rule {
+		regex = "giantswarm_observability_tenant"
+		action = "labeldrop"
+	}
+
+	// We drop any log lines that do not content a tenant id label
```
QuentinBisson, line 104: This is to make sure we have a tenant and not accept any logs that do not

pkg/resource/logging-config/alloy/logging.alloy.template
```diff
@@ -89,7 +89,25 @@ discovery.relabel "kubernetes_pods" {
...
+		regex = "giantswarm_observability_tenant"
+		action = "labeldrop"
+	}
+
+	// We drop any log lines that do not contain a tenant_id label
+	rule {
+		source_labels = ["__meta_kubernetes_pod_container_name", "__tenant_id__"]
+		regex = "(.+);"
+		action = "drop"
+	}
```
TheoBrigitte, line 109: Do we really need that ? I would assume that those would be rejected. This would make those silently fails in this case.

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
```diff
@@ -83,6 +83,26 @@ podLogs:
...
         - {{ . }}
         {{- end }}
+    relabelings:
+    - action: replace
+      targetLabel: "giantswarm_observability_tenant"
+      replacement: {{ .DefaultWriteTenant }}
+- name: customers-logs
+  namespace: kube-system
+  spec:
+    selector: {}
```
TheoBrigitte, line 93: I see we are selecting every PodLogs here, I am worried we run into cases where customers wants to run their own logging pipeline and suddenly they get their logs going through our infra as a side unwanted effect.
Using an explicit label to configure those PodLogs to selected by Giantswarm infrastructure would make it safer and more declarative.

pkg/resource/logging-config/alloy/logging.alloy.template
QuentinBisson: An error metric paging is not silent :)

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
QuentinBisson: What do you think about this then?

```suggestion
    selector:
      matchExpressions:
      - key: observability.giantswarm.io/tenant
        operator: Exists
```

pkg/resource/logging-config/alloy/logging.alloy.template
QuentinBisson: This will create a lot of warn logs in alloy as well as a lot of 401 in the loki gateway and nginx which is not really good imo. I prefer that we drop to avoid those

TheoBrigitte: ok why not

pkg/resource/logging-config/alloy/logging-config.alloy.yaml.template
TheoBrigitte: That works

TheoBrigitte (approved):

PR #301, opened 2025-01-16
fix workload cluster pod logs selectors

QuentinBisson:

This fixes the pod logs selectors as they customer WCs pod logs selectors were incorrectly exchanged

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


QuantumEnigmaa:
I'm not sure to understand what moving the logs selectors from `default-namespaces` to `customer-logs` changes :thinking:

QuentinBisson:
Well,  the one with default namespace should select all pods that are in the defaut namespaces while customer-logs should only select the pods that have a label

QuantumEnigmaa:
Oh ok it makes sense then :)

QuantumEnigmaa (approved):

PR #308, opened 2025-01-30
Fix `job` label in logs

TheoBrigitte:
Towards: https://github.com/giantswarm/roadmap/issues/3855

This PRs fixes the `job` label in the logs by removing the `job` relabeling rule which make the default job label appear in logs.

The job label will then be shown as `<pod log namespace/<pod log name>`.

image


QuentinBisson:
LGTM, do you think we could (at least for now) extract the pod and namespace label from the instance label?

PR #309, opened 2025-01-30
Fix namespace, pod, and container log label

TheoBrigitte:
Towards: https://github.com/giantswarm/roadmap/issues/3855

This PR fixes the namespace, pod, and container label shown in logs.

It does so by extracting those values from the `instance` label which is in the form of `<namespace>/<pod>:<container>`.

image

### DIFF
--- a/pkg/resource/logging-config/alloy/logging.alloy.template
+++ b/pkg/resource/logging-config/alloy/logging.alloy.template
@@ -68,6 +68,7 @@ discovery.relabel "kubernetes_pods" {
 		target_label  = "node"
 	}
 
+	{{- if .SupportPodLogs }}
 	rule {
 		source_labels = ["instance"]
 		regex         = "(.+)/.+"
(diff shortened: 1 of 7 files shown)


pkg/resource/logging-config/alloy/logging.alloy.template
```diff
@@ -69,17 +69,20 @@ discovery.relabel "kubernetes_pods" {
 	}
 
 	rule {
-		source_labels = ["__meta_kubernetes_namespace"]
+		source_labels = ["instance"]
```
QuentinBisson, line 72: I think we need to handle that differently as this part contains both discovery.relabel that predates pod logs and loki.relabel the future right?

TheoBrigitte: Yeah right, we need to support both versions

TheoBrigitte: Added a condition and use the new approach only for podlogs

QuentinBisson: Nice. Do you think it would not make sense to add the condition to the whole `discovery/loki.relabel` sections as discovery targets prepodlogs and the other targets pod logs? Otherwise i'm good

QuentinBisson (approved):

PR #324, opened 2025-02-21
Add tenant filtering in Alloy configuration

TheoBrigitte:
Towards: https://github.com/giantswarm/roadmap/issues/3885

This PR adds log tenant filtering in Alloy configuration.

It uses a regex to match against a list of tenants, log line with tenant not in the list will be dropped.


pkg/resource/logging-config/alloy/test/logging-config.alloy.170_WC_custom_tenants.yaml
```diff
@@ -0,0 +1,353 @@
...
+    clustering:
+      enabled: true
+      name: alloy-logs
+    extraEnv:
+    - name: NODENAME
+      valueFrom:
+        fieldRef:
+          fieldPath: spec.nodeName
+    mounts:
+      varlog: true
```
QuentinBisson, line 315: Why did this file change fully?

TheoBrigitte: That's a new file for a new test case with multiple tenants

pkg/resource/logging-config/alloy/logging.alloy.template
```diff
@@ -106,6 +106,7 @@ discovery.relabel "kubernetes_pods" {
 	{{- if .SupportPodLogs }}
 	rule {
 		source_labels = ["giantswarm_observability_tenant"]
+		regex         = "({{ join "|" .Tenants }})"
```
QuentinBisson, line 109: I'm not sure a fond of this regex as it's not stating why we are dropping the logs 🤔

pkg/resource/logging-config/reconciler.go
```diff
@@ -41,8 +41,11 @@ func (r *Reconciler) ReconcileCreate(ctx context.Context, lc loggedcluster.Inter
 		return ctrl.Result{}, errors.WithStack(err)
 	}
 
+	// TODO: implement tenants discovery
```
QuentinBisson, line 44: Well it's missing 😂

pkg/resource/logging-config/alloy/logging.alloy.template
TheoBrigitte: Do you mean we should add a comment here ?

pkg/resource/logging-config/reconciler.go
QuantumEnigmaa: This will be implemented in another PR

pkg/resource/logging-config/alloy/logging.alloy.template
QuentinBisson: Oh sorry, I think we should rather use https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagedrop instead to be able to set the `drop_counter_reason` then it would make it easy to craft an alert around it instead of trying to figure out why the logs are not ingested. It would be better UX wise IMO

TheoBrigitte: That would be great, I'll try this.

pkg/resource/logging-config/logging-config.go
```diff
@@ -61,3 +65,21 @@ func ConfigMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
 func getLoggingConfigName(lc loggedcluster.Interface) string {
 	return fmt.Sprintf("%s-%s", lc.GetClusterName(), loggingConfigName)
 }
+
+func listTenants(k8sClient client.Client, ctx context.Context) ([]string, error) {
+	tenantList := make([]string, 0)
```
QuentinBisson, line 70:
```suggestion
	tenants := make([]string, 0)
```

pkg/resource/logging-config/logging-config.go
```diff
@@ -61,3 +65,21 @@ func ConfigMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
...
+	var grafaOrgList v1alpha1.GrafanaOrganizationList
```
QuentinBisson, line 71:
```suggestion
	var grafanaOrganizations v1alpha1.GrafanaOrganizationList
```

pkg/resource/logging-config/logging-config.go
```diff
@@ -61,3 +65,21 @@ func ConfigMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
...
+
+	err := k8sClient.List(ctx, &grafaOrgList)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, grafanaOrg := range grafaOrgList.Items {
```
QuentinBisson, line 78:
```suggestion
	for _, organization := range grafaOrgList.Items {
```

pkg/resource/logging-config/logging-config.go
```diff
@@ -61,3 +65,21 @@ func ConfigMeta(lc loggedcluster.Interface) metav1.ObjectMeta {
...
+		for _, tenant := range grafanaOrg.Spec.Tenants {
+			tenantList = append(tenantList, string(tenant))
```
QuentinBisson, line 80: Can we make sure this list is unique? Maybe we do not care 🤔

QuantumEnigmaa: :heavy_check_mark:

PR #332, opened 2025-02-26
add logic to the grafana organization reconciler

QuantumEnigmaa:

Towards https://github.com/giantswarm/roadmap/issues/3885

This PR adds the logic to the grafana organization reconciler so that it triggers the logging config reconciler whenever there's a change in a grafana organization CR.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add tenant filtering in Alloy config
-- Add logic into `grafana-organization-reconciler`.
+- Add logic into `grafana-organization-reconciler` for updating the logging-config's list of tenants whenever a change occurs to the Grafana Organization CRs.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add tenant filtering in Alloy config
 - Add logic into `grafana-organization-reconciler`.
 
+### Changed
+
+- Use smaller dockerfile to reduce build time as ABS already generates the go binary.
+
 ### Fixed
 
 - Fix non-working log lines dropping on missing tenant id
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add tenant filtering in Alloy config
-- Add logic into `grafana-organization-reconciler` for updating the logging-config's list of tenants whenever a change occurs to the Grafana Organization CRs.
+- Implement `grafana-organization` controller to update tenant IDs filter on GrafanaOrganization changes.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add tenant filtering in Alloy config
+- Add logic into `grafana-organization-reconciler`.
 
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


QuantumEnigmaa:
Testing on `golem` and it currently works for the create and update use case but not for the delete one.

QuantumEnigmaa:
Tested on `golem` and `glean` and working for all use cases.

QuentinBisson:
As discussed during our pairing, I still prefer a watch over a full fledge controller because it is a lot less code but I'll let the others give their idea.

internal/controller/grafana_organization_reconciler.go
```diff
@@ -0,0 +1,96 @@
...
+	logger := log.FromContext(ctx)
+
+	logger.Info("Started reconciling Grafana Organization")
+	defer logger.Info("Finished reconciling Grafana Organization")
+
+	grafanaOrganization := &grafanaorganization.GrafanaOrganization{}
+	err = g.Client.Get(ctx, req.NamespacedName, grafanaOrganization)
+	if client.IgnoreNotFound(err) != nil { // Only handle errors that are not notFound ones
+		return ctrl.Result{}, errors.WithStack(err)
+	}
```
TheoBrigitte, line 64: This code can be removed, `grafanaOrganization` is not used.

internal/controller/grafana_organization_reconciler.go
```diff
@@ -0,0 +1,96 @@
...
+
+	for _, cluster := range clusters.Items {
+		loggedCluster := &capicluster.Object{
+			Object:  &cluster,
+			Options: loggedcluster.O,
+		}
+
+		// Reconcile logging config for each cluster
+		result, err := g.LoggingConfigReconciler.ReconcileCreate(ctx, loggedCluster)
```
TheoBrigitte, line 82: We should not call `ReconcileCreate` on cluster objects being deleted or when logging is disabled, please use `common.IsLoggingEnabled`, see https://github.com/giantswarm/logging-operator/blob/01a805e8397eca3bf44b04cdcbf0ab1a5428b399/pkg/logging-reconciler/logging_reconciler.go#L28-L32

internal/controller/grafana_organization_reconciler.go
```diff
@@ -0,0 +1,96 @@
...
+
+	loggedcluster "github.com/giantswarm/logging-operator/pkg/logged-cluster"
+	"github.com/giantswarm/logging-operator/pkg/logged-cluster/capicluster"
+	loggingconfig "github.com/giantswarm/logging-operator/pkg/resource/logging-config"
+)
+
+// GrafanaOrganizationReconciler reconciles grafanaOrganization CRs
+type GrafanaOrganizationReconciler struct {
+	client.Client
+	Scheme                  *runtime.Scheme
```
TheoBrigitte, line 39: Can we remove this ?

TheoBrigitte (commented):
We might also want to exclude GrafanaOrganization being deleted from within [`listTenants`](https://github.com/giantswarm/logging-operator/blob/01a805e8397eca3bf44b04cdcbf0ab1a5428b399/pkg/resource/logging-config/logging-config.go#L70)

TheoBrigitte:
> As discussed during our pairing, I still prefer a watch over a full fledge controller because it is a lot less code but I'll let the others give their idea.

An additional watch on the Cluster controller would run all the resources configured in the controller, which is a bit much given that we just want to update the Alloy config here. I prefer having a separate controller which just does this one thing.

internal/controller/grafana_organization_reconciler.go
QuantumEnigmaa: Yeah you' re right. 
Done :heavy_check_mark:

internal/controller/grafana_organization_reconciler.go
QuantumEnigmaa: Done :heavy_check_mark:

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add tenant filtering in Alloy config
+- Add logic into `grafana-organization-reconciler` for updating the logging-config's list of tenants whenever a change occurs to the Grafana Organization CRs.
```
TheoBrigitte, line 13:
```suggestion
- Implement `grafana-organization` controller to update tenant IDs filter on GrafanaOrganization changes.
```

TheoBrigitte (approved):

PR #333, opened 2025-02-26
simplify-dockerfile

QuentinBisson:

This PR is an attempt to use a simpler dockerfile to reduce operator build time as we in fact generate the binary twice (once with abs and once within the docker container)

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,33 +1,9 @@
-# Build the logging-operator binary
-FROM golang:1.24 as builder
-ARG TARGETOS
-ARG TARGETARCH
+FROM gsoci.azurecr.io/giantswarm/alpine:3.20.3-giantswarm
 
-WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
-
-# Copy the go source
-COPY main.go main.go
-COPY internal/ internal/
-COPY pkg/ pkg/
-
-# Build
-# the GOARCH has not a default value to allow the binary be built according to the host where the command
-# was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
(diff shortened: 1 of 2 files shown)


QuentinBisson:
PR was tested and deployed on golem, will add this to our other 2 operators

Dockerfile
```diff
@@ -1,33 +1,9 @@
...
-# the GOARCH has not a default value to allow the binary be built according to the host where the command
-# was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
-# the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
-# by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o logging-operator main.go
+ADD logging-operator logging-operator
 
-# Use distroless as minimal base image to package the logging-operator binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
```
QuantumEnigmaa, line 28: Why change the base image from `gcr.io/distroless/static:nonroot` to `gsoci.azurecr.io/giantswarm/alpine:3.20.3-giantswarm` ? 

If I'm not mistaken, the first one is smaller than the second right ?

hervenicol:
I understand why we would remove the build container because our pipeline already builds it (BTW it'd rather have it the other way around: build an image and then test it, but we have to comply with what ABS does).
But, like Zirko, I don't understand why the changes in the end-result container.

Dockerfile
QuentinBisson: I tool this from the ones generated by our tooling but I can check this one yes

Dockerfile
```diff
@@ -1,33 +1,9 @@
...
-WORKDIR /
```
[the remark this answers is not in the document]
QuentinBisson: It's not removed though but I added an extra line which I'm removing

QuentinBisson:
I addressed your reviews and will do so in the other 2 operators.

I agree Herve but I'm not sure I want to fight the process on this today :D

hervenicol (approved):
LGTM

PR #341, opened 2025-03-06
Only fetch tenant IDs on CAPI clusters

TheoBrigitte:
Fix the following error due to non-existing GrafanaOrganization CRD on vintage installations

```
no matches for kind "GrafanaOrganization" in version "observability.giantswarm.io/v1alpha1"
```


QuantumEnigmaa (approved):

TheoBrigitte:
Tested on vintage :heavy_check_mark:

PR #424, opened 2025-06-24
Refactor global options into config

QuentinBisson:

This pull requests gets rid of the global Options in the logged-cluster interface in favor of dependency injection to be able to eventually get rid of the logged-cluster interface in favor of cluster api type

### Checklist


main.go
```diff
@@ -157,6 +157,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Create Config for dependency injection
+	appConfig := &config.Config{
```
QuentinBisson, line 161:
```suggestion
	appConfig := config.Config{
```

pkg/reconciler/logging/logging_reconciler.go
```diff
@@ -24,10 +25,11 @@ type LoggingReconciler struct {
 	client.Client
 	Scheme      *runtime.Scheme
 	Reconcilers []reconciler.Interface
+	Config      *config.Config
```
QuentinBisson, line 28:
```suggestion
	Config      config.Config
```

QuentinBisson:
I understand and this is for now, a temporary solution because we do not want to use a global variable here but a passed down config object (same as we did in the observability operator).

The issue with that is that we need some kind of service layer and the logging operator is currently way too far from this with the reconciler architecture but yes this is my goal. 

This PR is only the beginning to go to that point eventually so we use the same architecture as observability operator so we can merge the 2 without blowing up the changes in the pull requests so you can actually review them :D

TheoBrigitte:
> I understand and this is for now, a temporary solution because we do not want to use a global variable here but a passed down config object (same as we did in the observability operator).
> 
> The issue with that is that we need some kind of service layer and the logging operator is currently way too far from this with the reconciler architecture but yes this is my goal.
> 
> This PR is only the beginning to go to that point eventually so we use the same architecture as observability operator so we can merge the 2 without blowing up the changes in the pull requests so you can actually review them :D

I don't know what you mean with service-layer.
In order to merge the 2 operator I think we should have a better discussion on how we want to achieve that without too much friction and less effort.

QuentinBisson:
In the observability operator, we use what's called a service layer (https://github.com/giantswarm/observability-operator/blob/main/pkg/bundle/service.go), which is where we store our business logic, what the reconciler actually does like configuring the bundle, the mimir ingress and so on in separate services (so objects with their own configs and so on)

In the logging-operator, we use the old operatorkit resource or reconciler pattern, that uses public functions and not functions scoped to the reconciler like this [GenerateLoggingSecret](https://github.com/giantswarm/logging-operator/blob/75777af4ee6e8262aa982a0f46cd92bb88771eac/pkg/resource/logging-secret/logging-secret.go#L19C6-L19C27) so we do not have access to the reconciler itself.

To facilitate the merge, we will need to follow the same pattern (create services) and adding the ocnfig in the public non scoped function is the first step, later on, we can think about transforming the reconciler into services with the config injected

PR #425, opened 2025-06-24
clean up logged-cluster interface again

QuentinBisson:
### What this PR does / why we need it

Follow up of https://github.com/giantswarm/logging-operator/pull/424 to clean up logging operator even further

### Checklist

- [x] Update changelog in CHANGELOG.md.
- [ ] Make sure the new features are scoped to supported observability-bundle versions (see `supportPodLogs` boolean as an example).
- [ ] Follow deployment test procedure in the tests/manual_e2e directory and have a working branch.

PR #427, opened 2025-06-24
Clean up

QuentinBisson:

Clean up reconciler interface in favor of using the controller

THis is a follow up of Follow up of https://github.com/giantswarm/logging-operator/pull/424 and https://github.com/giantswarm/logging-operator/pull/425 and https://github.com/giantswarm/logging-operator/pull/426

### Checklist

- [ ] Update changelog in CHANGELOG.md.


pkg/resource/interface.go
```diff
@@ -1,4 +1,4 @@
-package reconciler
+package resource
```
hervenicol, line 1: Why renaming it from `reconciler` to `resource`?
Especially since all instantiations are called `Reconciler`, and comments are still talking about `reconciler`.

QuentinBisson: Because the package is actually pkg/resource :D

hervenicol: > Because the package is actually pkg/resource :D

It looks to me that it was not before this PR, and it has been renamed right here in this PR. :thinking: 
image

hervenicol: I don't remember this discussion with Theo.
And I don't know what are yo services.

The base reason for my question was that:
- the interface changed package and is now `resource` instead of `reconciler`
- but the code and comments are still talking about `reconciler` in lots of places, which can be confusing.

I'm sorry, I feel like I'm saying nonsense and I can't get understood + I can't understand your answers :rofl: 
I'll assume for the best, you know what you're doing even though I don't :sweat_smile:

QuentinBisson: Thanks :)
The discussion was in another PR, sorry. I understand what you mean though. What I want to say is that in the observability operator we have a service/domain layer with the grafana service, Mimir service, bundle configuration service and here we have reconcilers. I plan on moving the logging operator to the same code architecture so we can move services from one operator to the next. That will required removing the reconciler interface anyway.

Now the current code will still talk about réconciliées, and this is not fine so i'll update those next week

QuentinBisson:
@hervenicol That's a lot more changes but it's more accurate

PR #450, opened 2025-08-18
Update tests

hervenicol:
### What this PR does / why we need it

The latest CI jobs (https://github.com/giantswarm/logging-operator/pull/448) looks for trailing spaces.
This PR removes them, including useless empty lines in generated test config files.

### Checklist

- [x] Update changelog in CHANGELOG.md.
- [ ] Make sure the new features are scoped to supported observability-bundle versions (see `supportPodLogs` boolean as an example).
- [ ] Follow deployment test procedure in the tests/manual_e2e directory and have a working branch.

PR #451, opened 2025-08-18
Add make update-golden-files target

TheoBrigitte:
Adding a `make update-golden-files` target to the Makefile to easily update the golden files used in tests.


hervenicol (commented):
Nice!

That would be great to also add a word about it in the README, but it would conflict with what I added in this PR: https://github.com/giantswarm/logging-operator/pull/450

hervenicol (approved):
The readme now already mentions it, so let's merge :rocket:

PR #465, opened 2025-09-02
Add priorityclass alloy logs

QuantumEnigmaa:
### Checklist


QuentinBisson:
I don't mind this approach but I do not think we want alloy logs to have the same priority as the alloy metrics though right? We should use a less priority one

QuentinBisson:
Then let's try

